PR #14, opened 2023-11-10
Update ParseNodeIDs to use ExpandedNodeID instead of NodeID

### LINKED ISSUE #5: Using Expanded-NodeId-Strings rather then regular NodeId-String in configuration
### Example:
NodeId -> `ns=2;s=IoTSensors`
Expanded-NodeId -> `nsu=http://myIoTSensorNamspace/UA;s=IoTSensors`

In scenarios in which changes happen on the shopfloor (OPC UA Server) there is the chance that the NamespaceIndex changes and your configuration will be broken. 

Using the NamespaceIndex in a configuration is not a good practice, you should consider switch to the Expanded-NodeId format which uses the NamespaceUri instead of the index. In order to leverage the benefits of the expanded format the OPC UA Client needs to:
1. Connect
2. read the NamespaceArray
3. convert the Expanded-NodeIds once to the correct Index (verification that, i am on the correct server with the correct namespaces, which i would expect)
4. continue with reading/subscribing/...

this would lead to a much more robust behavior of the OPC UA Client. And if something changes e.g. someone imports a UA-Model or a CS your config will still be valid. This is a issue most commercial products get wrong in implementation!

devantler:
This pull request updates the `ParseNodeIDs` function to use `ExpandedNodeID` instead of `NodeID`. The function now takes in a slice of strings and returns a slice of `ExpandedNodeID` pointers. Additionally, the function now parses the `ExpandedNodeID` using the `ParseExpandedNodeID` function and a `NamespaceArray` obtained from the server. This change ensures that the `ParseNodeIDs` function is up-to-date with the latest OPC-UA specifications.

Fixes #5

### DIFF
--- a/plugin/opcua.go
+++ b/plugin/opcua.go
@@ -190,14 +190,17 @@ var OPCUAConfigSpec = service.NewConfigSpec().
 	Field(service.NewStringField("password").Description("The password to connect to the server. Defaults to none.").Default("")).
 	Field(service.NewStringListField("nodeIDs").Description("The OPC-UA node IDs to start the browsing."))
 
-func ParseNodeIDs(incomingNodes []string) []*ua.NodeID {
+func ParseNodeIDs(incomingNodes []string) []*ua.ExpandedNodeID {
 
 	// Parse all nodeIDs to validate them.
 	// loop through all nodeIDs, parse them and put them into a slice
-	parsedNodeIDs := make([]*ua.NodeID, len(incomingNodes))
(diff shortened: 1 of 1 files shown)


devantler:
I have to abandon this PR. We in my will not strictly require the functionality, so I am, under company policy, not allowed to contribute to this specific issue.

For sharing what I learned prior to this decision, I believe it should be possible to retrieve the node namespace array by sending a request to "i=2255", and then use this array where the ExpandedNodeIds are resolved. Otherwise, I believe it is just an API change as demonstrated in these early changes.

PR #29, opened 2024-02-01
Added S7comm read plugin

JeremyTheocharis:
All tests pass


.github/workflows/s7.yml
```diff
@@ -0,0 +1,44 @@
...
+  push:
+    branches:
+      - '**'
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  GO_VERSION: '1.21.*'
+
+concurrency:
+  group: s7-test
```
Scarjit, line 28: I assume you want to run wago and s7 in parallel ?

Scarjit (approved):

.github/workflows/s7.yml
JeremyTheocharis: yes

s7comm_plugin/s7comm_test.go
```diff
@@ -0,0 +1,166 @@
...
+			tcpDevice:    endpoint,
+			rack:         rack,
+			slot:         slot,
+			batchMaxSize: batchMaxSize,
+			batches:      batches,
+		}
+
+		// Attempt to connect
+		err = input.Connect(ctx)
+		defer input.Close(ctx)
```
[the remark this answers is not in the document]
JeremyTheocharis: That is a fair point. However, knowing the reliability of the PLCs I would rather have
1. a connection being too closed too often than
2. a connection being left open, causing the PLC to crash

Therefore I would leave it like this, even though it might not be the best from a coding approach (closing a connection that failed to initialize).

PR #34, opened 2024-02-20
Use NodeID instead of BrowseName in Path in case of duplicate names

DaruZero:
see https://link.excalidraw.com/l/3xbwAQtsCRd/3gVKaxIqv0T

### DIFF
--- a/opcua_plugin/opcua.go
+++ b/opcua_plugin/opcua.go
@@ -497,30 +497,7 @@ func (g *OPCUAInput) Connect(ctx context.Context) error {
 			return err
 		}
 
-		// cycle through all nodes and check if two nodes have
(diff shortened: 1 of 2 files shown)


opcua_plugin/opcua.go
```diff
@@ -492,6 +497,31 @@ func (g *OPCUAInput) Connect(ctx context.Context) error {
 			return err
 		}
 
+		// cycle through all nodes and check if two nodes have
```
JeremyTheocharis, line 500: This is perfect for unit testing. Can you add a unit test there?

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #35, opened 2024-03-02
added ads plugin and readme

RuneRoven:
Added ads plugin and changed readme. Import for new package still need to be changed in the main file. Import for ads native package is local so that might need to be changed.


ads_plugin/nativeADS/readWriter.go
```diff
@@ -0,0 +1,341 @@
...
+			}
+			i := binary.LittleEndian.Uint32(data[start:stop])
+			t := time.Unix(0, int64(uint64(i)*uint64(time.Second))-int64(time.Hour))
+
+			newValue = t.Truncate(time.Millisecond).Format("2006-01-02 15:04:05")
+		default:
+			return "", fmt.Errorf("unknown format cannot parse")
+		}
+	}
+	if strcmp(symbol.Value, newValue) != 0 &&
```
Scarjit, line 150: Is there a reason why you don't compare the strings directly ?
e.g: `if symbol.Value != newValue`

ads_plugin/nativeADS/readWriter.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+func (symbol *Symbol) parentChanged() {
+	if symbol.Parent != nil {
+		symbol.Parent.parentChanged()
+	}
+	symbol.Changed = true
+}
+
+func stringArrayIncludes(vs []string, t string) bool {
```
Scarjit, line 171: https://pkg.go.dev/golang.org/x/exp/slices#Contains would be more idiomatic

ads_plugin/nativeADS/readWriter.go
RuneRoven: That's from the native ads library which is not my code originally, so i have now idea

ads_plugin/nativeADS/readWriter.go
RuneRoven: That's from the native ads library which is not my code originally, so i have now idea why..

RuneRoven (commented):
good clean up. you made some changes in the symbol reading with the offset? Why? also in the test suite, make sure the ams net id on the target is correct, it is not always the same as the ip + .1.1 however on the client side this is usually the case

JeremyTheocharis:
> good clean up. you made some changes in the symbol reading with the offset? Why? also in the test suite, make sure the ams net id on the target is correct, it is not always the same as the ip + .1.1 however on the client side this is usually the case

The offset was actually never used, and I got a warning from the go static checker that the variable was never used. So I removed it. Maybe it indicates a logic error of some sort, that an offset should be used but was not used?

Still need to write the test fully

JeremyTheocharis:
We have now merged it using a different method :)

PR #56, opened 2024-07-15
added opcua source and server timestamp

trarbr:
This PR adds [SourceTimestamp](https://reference.opcfoundation.org/Core/Part4/v104/docs/7.7.3) and [ServerTimestamp](https://reference.opcfoundation.org/Core/Part4/v104/docs/7.7.4) to the metadata of the Benthos message. Having these timestamps is very useful in situations where the data passes through multiple PLC's before reaching Benthos.

### DIFF
--- a/opcua_plugin/opcua.go
+++ b/opcua_plugin/opcua.go
@@ -414,7 +414,8 @@ func UpdateNodePaths(nodes []NodeDef) {
 
 // createMessageFromValue creates a benthos messages from a given variant and nodeID
 // theoretically nodeID can be extracted from variant, but not in all cases (e.g., when subscribing), so it it left to the calling function
-func (g *OPCUAInput) createMessageFromValue(variant *ua.Variant, nodeDef NodeDef) *service.Message {
+func (g *OPCUAInput) createMessageFromValue(dataValue *ua.DataValue, nodeDef NodeDef) *service.Message {
+	variant := dataValue.Value
 	if variant == nil {
 		g.Log.Errorf("Variant is nil")
 		return nil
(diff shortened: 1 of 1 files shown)


JeremyTheocharis:
Nice and Thank you! Can you sign the CLA so we can accept your PR?

opcua_plugin/opcua.go
```diff
@@ -634,7 +637,7 @@ func (g *OPCUAInput) ReadBatchSubscribe(ctx context.Context) (service.MessageBat
 				handleID := item.ClientHandle
 
 				if uint32(len(g.NodeList)) >= handleID {
-					message := g.createMessageFromValue(item.Value.Value, g.NodeList[handleID])
+					message := g.createMessageFromValue(item.Value, g.NodeList[handleID])
```
JeremyTheocharis, line 640: May I ask for what the change here is?

opcua_plugin/opcua.go
```diff
@@ -488,6 +489,8 @@ func (g *OPCUAInput) createMessageFromValue(variant *ua.Variant, nodeDef NodeDef
 	message.MetaSet("opcua_path", sanitize(nodeDef.NodeID.String()))
 	message.MetaSet("opcua_tag_path", sanitize(nodeDef.BrowseName))
 	message.MetaSet("opcua_parent_path", sanitize(nodeDef.ParentNodeID))
+	message.MetaSet("opcua_source_timestamp", dataValue.SourceTimestamp.Format("2006-01-02T15:04:05.000000Z07:00"))
```
JeremyTheocharis, line 492: happy to add those!

opcua_plugin/opcua.go
trarbr: Sure. The timestamps that I need are properties of the DataValue struct, which is what `item.Value` points to. `item.Value.Value` points to a Variant, which is what the current `createMessageFromValue` function works with. I just changed the `createMessageFromValue` function to accept the DataValue instead, so it has access to both the timestamps and the Variant.

trarbr:
> Nice and Thank you! Can you sign the CLA so we can accept your PR?

Sorry for the delay, I just need to work this out with my employer. I will let you know when it has been resolved.

trarbr:
@JeremyTheocharis The CLA was approved and I have now signed it

JeremyTheocharis (approved):

PR #63, opened 2024-09-30
ENG update redpanda connec to 4.37.0

Scarjit:
Sadly redpanda only created a release but did not bump the version in the bundle


JeremyTheocharis (approved):

Scarjit:
Not sure if the test is broken/inconsitent here, or if the update really broke opc-ua.
@JeremyTheocharis Any ideas ?

PR #64, opened 2024-10-08
[ENG-1569] Fix Ginkgo tests

kanapuli:
## Descritpion 

- Eventually is used in the tests to Handle ReadBatch

### DIFF
--- a/opcua_plugin/opcua.go
+++ b/opcua_plugin/opcua.go
@@ -27,7 +27,7 @@ import (
 )
 
 const SessionTimeout = 5 * time.Second
-const SubscribeTimeoutContext = 10 * time.Second
+const SubscribeTimeoutContext = 3 * time.Second
 
 var OPCUAConfigSpec = service.NewConfigSpec().
 	Summary("Creates an input that reads data from OPC-UA servers. Created & maintained by the United Manufacturing Hub. About us: www.umh.app").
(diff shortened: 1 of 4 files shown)


opcua_plugin/opcua.go
```diff
@@ -219,9 +220,16 @@ func (g *OPCUAInput) Connect(ctx context.Context) (err error) {
...
+		g.SubNotifyChan = make(chan *opcua.PublishNotificationData, 10000)
+	}
 	// Browse and subscribe to the nodes if needed
 	// Do this asynchronously so that the first messages can already arrive
+	var wg sync.WaitGroup
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		g.Log.Infof("Please note that browsing large node trees can take some time")
 		if err := g.BrowseAndSubscribeIfNeeded(ctx); err != nil {
```
kanapuli, line 234: **Note for Reviewers:** This code is somewhat complex. The function `g.BrowseAndSubscribeIfNeeded` must be completed before the `Connect` function, as it initializes `g.NodeList`. This `NodeList` is necessary for the subsequent `ReadBatch` function. If `g.BrowseAndSubscribeIfNeeded` is executed without using `sync.WaitGroup`, there is no assurance that the goroutine will finish before `ReadBatch` is called. This is a primary reason for test failures, as the goroutine may not complete, leading to an empty `NodeList` and causing `ReadBatch` to fail.

If I remove the `sync.WaitGroup` from this section and run `g.BrowseAndSubscribeIfNeeded` synchronously, the tests still fail due to some channels becoming blocked. Therefore, this section needs to be refactored in the future for improved simplicity and clarity.

opcua_plugin/opcua.go
```diff
@@ -219,9 +220,16 @@ func (g *OPCUAInput) Connect(ctx context.Context) (err error) {
...
+	// Create a subscription channel if needed
+	if g.SubscribeEnabled {
+		g.SubNotifyChan = make(chan *opcua.PublishNotificationData, 10000)
+	}
 	// Browse and subscribe to the nodes if needed
 	// Do this asynchronously so that the first messages can already arrive
+	var wg sync.WaitGroup
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
```
JeremyTheocharis, line 232: then you can also instead remove the goroutine. would make the code a little bit cleaner

JeremyTheocharis (changes requested):

opcua_plugin/opcua.go
```diff
@@ -248,6 +243,12 @@ func (g *OPCUAInput) Connect(ctx context.Context) (err error) {
 // The function updates heartbeat information and monitors the connection's health.
 // If no messages or heartbeats are received within the expected timeframe, it closes the connection.
 func (g *OPCUAInput) ReadBatch(ctx context.Context) (msgs service.MessageBatch, ackFunc service.AckFunc, err error) {
+	if len(g.NodeList) == 0 {
```
JeremyTheocharis, line 246: I am not a fan of this, but I can understand that it is needed for tests. Can we ensure that this never happens in production, where connect is called before ReadBatch? For example that at the end of connect we check if the nodelist is empty and if yes, disconnect again abnd return an error so that benthos tries to re-connect a couple of seconds later. then we need to update this code here to explain that this is just for tests.

Or we could also instead update all the tests so that the nodes are always there before ReadBatch is called

I am worried that this might lead to unintended side-effects like during conncet it does not find any nodes, but continues anyway. and that if the nodes later come online in the server, that they are here then suddenly browsed. However when writing about this, it actually makes sense to do it like this. Maybe still add a code comment about it?

JeremyTheocharis (approved):

PR #67, opened 2024-10-16
[ENG-1744] Modify GetNodes to return additional NodeID map

kanapuli:
## Description 

GetNodes is modified to return an additional map of nodeID to nodeName


JeremyTheocharis (changes requested):
Can you explain the "why" this exists and what for? also it would be better to call it mapping of the path ("root.objects.server") to the NodeID ("i=86)

JeremyTheocharis (approved):

PR #78, opened 2024-11-08
Added support for EIO404 Bluetooth Mesh IoT Base Station.

EmlingerM:
### Breaking Changes:

Instead of using only the port number as the unique identifier for connected devices, the entire URI will be used.

e.g. `string: /iolinkmaster/port[1]/` or `string: /meshnetwork/mesh_adapter[1]/iolinkmaster/port[1] `

This replaces the previous method of using eg. `int: 1` to identify a connected sensor.

## Release Notes

- **New Features**
  - Enhanced documentation in the README, including detailed descriptions for OPC UA, S7comm, Modbus, and SensorConnect plugins.
  - Added support for additional devices in the SensorConnect plugin.
  - Introduced a new section for Beckhoff ADS protocol.

- **Improvements**
  - Streamlined handling of connected devices, shifting from a port-centric to a device-centric approach.
  - Improved error handling and logging for better clarity and troubleshooting.

- **Documentation**
  - Expanded sections on Modbus, SensorConnect, and testing requirements for better user guidance.

### DIFF
--- a/sensorconnect_plugin/downloadPortModeData.go
+++ b/sensorconnect_plugin/downloadPortModeData.go
@@ -22,7 +22,7 @@ type ConnectedDeviceInfo struct {
 	Serial      string
 	Port        string
 	UseRawData  bool // If true, the raw data will be used instead of the parsed data. This flag is automatically set if there is no IODD file available
-	btAdapter   string
+	BtAdapter   string
 }
 
 // BadgeSize is Maximum possible size of the "/getdatamulti" request, determined for AL1352 and EIO404 Devices.
@@ -60,9 +60,9 @@ func (s *SensorConnectInput) GetConnectedDevices(ctx context.Context) ([]Connect
 		btAdapter, err := extractBluetoothAdapter(uri)
 		if err != nil {
 			s.logger.Debugf(err.Error())
-			deviceInfo.btAdapter = "none"
+			deviceInfo.BtAdapter = "none"
 		} else {
-			deviceInfo.btAdapter = btAdapter
+			deviceInfo.BtAdapter = btAdapter
 		}
 
 		port, err := extractPort(uri)
@@ -109,7 +109,7 @@ func (s *SensorConnectInput) GetConnectedDevices(ctx context.Context) ([]Connect
 			deviceIDValue, exists := data[deviceIDKey]
 			if !exists || deviceIDValue.Code != 200 || deviceIDValue.Data == nil {
 				if deviceIDValue.Code == 503 {
-					s.logger.Debugf("DeviceID not available for port %d (code 503), marking as disconnected", uri)
+					s.logger.Debugf("DeviceID not available for %s (code 503), marking as disconnected", uri)
 					deviceInfo.Connected = false
 					connectedDevices = append(connectedDevices, deviceInfo)
 					continue
@@ -179,7 +179,7 @@ func (s *SensorConnectInput) GetConnectedDevices(ctx context.Context) ([]Connect
 			serialKey := uri + "/iolinkdevice/serial"
 			serialValue, exists := data[serialKey]
 			if !exists || serialValue.Data == nil || serialValue.Code == 503 {
-				s.logger.Infof("Serial number not available for port %d", uri)
+				s.logger.Infof("Serial number not available for port %s", uri)
 			} else if serialValue.Code == 200 {
 				serial, err := parseString(serialValue.Data)
 				if err != nil {
--- a/sensorconnect_plugin/downloadPortModeData.go
+++ b/sensorconnect_plugin/downloadPortModeData.go
@@ -3,10 +3,12 @@ package sensorconnect_plugin
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"maps"
 	"regexp"
 	"strconv"
+	"strings"
 )
 
 // ConnectedDeviceInfo represents information about a connected device on a port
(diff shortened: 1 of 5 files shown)


EmlingerM:
Ok sorry for confusion.

This PR should not disrupt existing functionality. I have tested the code with both the EIO404 + EIO344  and AL1352 models, using various sensors.

The concept stays the same as before:

The logic remains consistent with previous implementations:

- When querying the portMode for "physical ports" on the EIO404 (e.g., /iolinkmaster/port[1]/), the EIO404 returns a status code indicating that the port does not exist or is not connected.
- Similarly, if you request the portMode for a Bluetooth Adapter  (e.g., /meshnetwork/mesh_adapter[1]/iolinkmaster/port[1]/), on a normal AL1352, it returns the same status code indicating that this no Bluetooth adapter are connected. 

This approach mirrors the previous behavior observed with the AL1350 (4 ports) and AL1352 (8 ports),


It would be possible change this logic to first query the specific model and then request only the portModes of the existing "ports" However, this would lead to a larger conceptual changes.

JeremyTheocharis:
alright, let me check it by looking into the code

also ENG-1909 (ignore this, just for our internal ticketing system)

sensorconnect_plugin/downloadPortModeData.go
```diff
@@ -210,33 +233,54 @@ func (s *SensorConnectInput) getUsedPortsAndMode(ctx context.Context) (RawUsedPo
...
 	}
 
-	requestData := map[string]interface{}{
-		"code": "request",
-		"adr":  "/getdatamulti",
-		"data": map[string]interface{}{
-			"datatosend": dataPoints,
-		},
+	// The EIO404 Bluetooth Mesh IoT Base Station supports up to 50 EIO344 Bluetooth Mesh IO-Link Adapters.
+	// Note that each EIO344 Bluetooth Mesh IO-Link Adapter is limited to a single IO-Link port.
```
JeremyTheocharis, line 237: Can you add here the note that you wrote me, that the EIO404 will ignore the `/iolinkmaster/port[%d]/mode` data points and vice versa? I think it would help here for understanding, that we expect to get a lot of 400 error codes back.

sensorconnect_plugin/downloadPortModeData.go
```diff
@@ -210,33 +233,54 @@ func (s *SensorConnectInput) getUsedPortsAndMode(ctx context.Context) (RawUsedPo
...
+			fmt.Sprintf("/meshnetwork/mesh_adapter[%d]/iolinkmaster/port[1]/iolinkdevice/serial", meshAdapter),
+		)
 	}
 
-	response, err := s.SendRequestToDevice(ctx, requestData)
-	if err != nil {
-		return RawUsedPortsAndMode{}, err
-	}
+	// To avoid HTTP 413 "Payload Too Large" or 507 "Insufficient Storage" errors, we need to request data in batches, splitting the payload into smaller, manageable chunks.
+	badgeSize := 50
```
JeremyTheocharis, line 249: why 50? maybe put this as constant in the header with a short explanation

sensorconnect_plugin/downloadPortModeData.go
```diff
@@ -210,33 +233,54 @@ func (s *SensorConnectInput) getUsedPortsAndMode(ctx context.Context) (RawUsedPo
...
-		s.logger.Errorf("Failed to marshal response: %v", err)
-		return RawUsedPortsAndMode{}, err
-	}
-	err = json.Unmarshal(responseBytes, &rawData)
-	if err != nil {
-		s.logger.Errorf("Failed to parse response: %v", err)
-		return RawUsedPortsAndMode{}, err
+		response, err := s.SendRequestToDevice(ctx, requestData)
+		if err != nil {
+			return result, err
```
JeremyTheocharis, line 264: if one of the requests fails, then it will send back the result array up to this moment, and an error message. but then we would have inconsistent amount of messages.

wouldn't it be better to continue here to let all the other requests execute as well. and then additionally return all errors? then if always the same port / message / batch fails ,all the others will continue to work

sensorconnect_plugin/downloadPortModeData.go
```diff
@@ -252,21 +296,37 @@ type UPAMDatum struct {
...
 	}
-	outputNumber, err := strconv.Atoi(outputSlice[0])
-	if err != nil {
-		err := fmt.Errorf("failed to convert string to integer: %s", outputSlice[0])
-		return -1, err
+
+	return "", fmt.Errorf("cannot find uri in %s", input)
+}
+
+func extractPort(input string) (string, error) {
```
JeremyTheocharis, line 310: can you put a quick function comment? a basic one from chatgpt is fine. it should answer why we need this function (not what the function does)

sensorconnect_plugin/downloadPortModeData.go
```diff
@@ -252,21 +296,37 @@ type UPAMDatum struct {
...
+
+	if len(matches) > 1 {
+		return matches[1], nil
 	}
-	return outputNumber, nil
+
+	return "", fmt.Errorf("cannot find port in %s", input)
+}
+
+func extractBluetoothAdapter(input string) (string, error) {
```
JeremyTheocharis, line 321: see comment for extractPort

JeremyTheocharis (changes requested):
Thank you for the PR! this is a nice addon, and tbh I did not even knew this hardware from ifm existed. 

I added some comments, mostly about comments and edge cases. I hope this does not discourage you from openeing more PRs. if yes, please tell em and I implement the changes myself.

sensorconnect_plugin/downloadPortModeData.go
EmlingerM: Yes good point. 👍 

I've changed this.

JeremyTheocharis:
We are currently trying out coderabbit, so I am going to let this "test" on your PR. No need to actually change something because of it, it is pure optional. But we saw it in other PRs that it was really helpful.

@coderabbitai

JeremyTheocharis (approved):

PR #102, opened 2025-01-14
ENG-2209: configure pipeline to use fallbacks

led0nk:
### Description:
- changed to go-version 1.23 since benthos-umh runs on 1.23
- install [tcping](https://github.com/cloverstd/tcping) (go wrapper for ping) there's an alternative workflow for port-check but i don't trust that
- check with tcping for either s7-main availability or fallback -> set output to true/false
- check with tcping for either wago-main availability or fallback -> set output to true/false
- depending on the output set env variable for test -> of nothing is available just quickly exit with failure

### Testing:
- endpoints were tested locally with setting them and Makefile
- workflows were tested with act-environment

## Summary by CodeRabbit

- **Workflow Updates**
	- Updated workflow names from "main" to project-specific names.
	- Upgraded Go version from 1.22.* to 1.23.0 across multiple workflows.
	- Enhanced testing workflows with port availability checks for S7 and Wago endpoints.

- **Documentation**
	- Updated README with specific model and firmware details for testing targets.
	- Added precise version information for WAGO PFC100, OPC-UA Server, and Siemens S7-1200.

- **Testing**
	- Minor import statement adjustments in test files.
	- Corrected typographical errors in test suite descriptions.

### DIFF
--- a/.github/workflows/opcua-plc.yml
+++ b/.github/workflows/opcua-plc.yml
@@ -99,19 +99,19 @@ jobs:
         run: |
           if [ "${{ steps.check_s7.outputs.available }}" == "true" ]; then
             echo "using s7 main"
-            TEST_S7_ENDPOINT_URI=${{ secrets.TEST_S7_ENDPOINT_URI }}
+            export TEST_S7_ENDPOINT_URI=${{ secrets.TEST_S7_ENDPOINT_URI }}
           elif  [ "${{ steps.check_s7_fallback.outputs.available }}" == "true" ]; then
             echo "using s7 fallback"
-            TEST_S7_ENDPOINT_URI=${{ secrets.TEST_S7_ENDPOINT_URI_FALLBACK }}
+            export TEST_S7_ENDPOINT_URI=${{ secrets.TEST_S7_ENDPOINT_URI_FALLBACK }}
           else
             exit 1
           fi
           if [ "${{ steps.check_wago.outputs.available }}" == "true" ]; then
             echo "using wago main"
-            TEST_WAGO_ENDPOINT_URI=${{ secrets.TEST_WAGO_ENDPOINT_URI }}
+            export TEST_WAGO_ENDPOINT_URI=${{ secrets.TEST_WAGO_ENDPOINT_URI }}
           elif [ "${{ steps.check_wago_fallback.outputs.available }}" == "true" ]; then
             echo "using wago fallback"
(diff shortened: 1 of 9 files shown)


.github/workflows/s7-plc.yml
```diff
@@ -42,5 +42,42 @@ jobs:
...
+          set +x
+          ENDPOINT="${{ secrets.TEST_S7_TCPDEVICE }}"
+          if tcping "$ENDPOINT" | grep -qi "Connected"; then
+            echo "available=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "available=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check S7-fallback port availability
+        id: check_s7_fallback
```
kanapuli, line 60: An optimal algorithm  could be add an if condition in this step (refer this https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution#overview) 

```
if: ${{ steps.check_s7.outputs.available}}
```

and set the $ENDPOINT to the fallback url. 

By this optimisation, this step won't run if the previous step is available and the step after this can be removed completely as the Endpoint is already set. 
This optimisation can be done in the other file too where you did a similar change.

.github/workflows/opcua-plc.yml
```diff
@@ -45,5 +45,76 @@ jobs:
...
         run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.19.0
+      - name: Install Tcping
+        run: go install github.com/cloverstd/tcping@latest
+      - name: Check S7 port availability
+        id: check_s7
+        run: |
+          set +x
+          URI="${{ secrets.TEST_S7_ENDPOINT_URI }}"
+          ENDPOINT="${URI#opc.tcp://}"
+          if tcping "$ENDPOINT" | grep -qi "Connected"; then
```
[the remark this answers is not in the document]
led0nk: Yes, i noticed it's default settings are a count of 4 and a timeout of 1s, so I applied that by flags. Because of this i noticed that the flags were changed slightly so i pinned the version to v0.1.1

.github/workflows/s7-plc.yml
led0nk: I have remodeled the workflow a bit regarding your comments.
We now use if-statements if the `main-device` is available, but it's not used for the testing-step since we want to exit & fail there if they're both not available.
On top of that we directly set the env-variable, so that we can check on it later on.

kanapuli (approved):

JeremyTheocharis (approved):

PR #104, opened 2025-01-14
BugFix: Consider only variable and reference classes

kanapuli:
## Description 

- While browsing the child nodes, the nodeclass type is currently passed as `ua.NodeClassVariable | ua.NodeClassObject`. This leads to a bug since the code tries to do a `bitwise OR` operation resulting to the value of `0x03`. But if you look at the enums of `NodeClass` there is no nodeclass with the value of `3`.  (See the nodeclass image below). This results in an incorrect browsing when `BrowseHierarchicalReferences is true`

image

## Image showing the result of Bitwise OR operation `ua.NodeclassVariable | ua.NodeClassObject`

image

## OPCUA Browser Records with the PR Change 

image

## OPCUA Browser Records before the Change

image

## Summary by CodeRabbit

- **Improvements**
	- Enhanced node browsing functionality with detailed logging for better observability.
	- Improved handling and logging of node attributes and references.
	- Added a new method for direct access to the underlying OPC UA node in the interface.
	- Extended mock structure to include a reference to the OPC UA node for testing purposes.

### DIFF
--- a/opcua_plugin/browse.go
+++ b/opcua_plugin/browse.go
@@ -131,7 +131,6 @@ func browse(ctx context.Context, n NodeBrowser, path string, level int, logger L
 		sendError(ctx, err, errChan, logger)
 		return
 	}
-	logger.Debugf("browseName for node %s is %s", n.ID().String(), browseName.Name)
 
 	var newPath string
 	if path == "" {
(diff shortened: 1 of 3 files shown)


JeremyTheocharis (changes requested):
Can you add a test for this behaviour? have you checked with mateusz whether this is actually the case?

PR #109, opened 2025-01-19
ENG-380: Use Extended-NodeID strings

led0nk:
I will need some advice here since this seems to work just fine by:
- connecting
- reading namespaceArray of node `i=2255`
- converting the namespaceArray-result to a usable `[]string`
- providing this to `ua.ParseExpandedNodeID` since this checks for the namespace-uri (If provided) and returning the related index to get a format like e.g. `<ns=4; s= i=1>`

benthos-expandednodeID
(I removed the Printf's for this draft)

What is still quite poor (imo):
- context-handling
- maybe at least some Info-logging to see if namespace was proven correctly here?
- should be only usable through feature-flag (currently)
- creating a new connection since it could be more intentional to move this to here:
https://github.com/united-manufacturing-hub/benthos-umh/blob/aa2177b55d0f455c830bfe80ae577050a0f960f5/opcua_plugin/server_info.go#L21-L112

### DIFF
--- a/opcua_plugin/opcua.go
+++ b/opcua_plugin/opcua.go
@@ -16,6 +16,7 @@ package opcua_plugin
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


opcua_plugin/opcua.go
```diff
@@ -67,6 +68,56 @@ func ParseNodeIDs(incomingNodes []string) []*ua.NodeID {
...
+func ReadNamespaceArray(endpoint string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
+
+	client, err := opcua.NewClient(endpoint, opcua.AuthAnonymous())
+	if err != nil {
+		return nil, err
+	}
+
+	err = client.Connect(ctx)
```
kanapuli, line 93: (major) You should be reusing the Client that is initialized from the connect method.

opcua_plugin/opcua.go
```diff
@@ -67,6 +68,56 @@ func ParseNodeIDs(incomingNodes []string) []*ua.NodeID {
...
+		if err != nil {
+			return nil, err
+		}
+		parsedExpandedNodeIDs = append(parsedExpandedNodeIDs, parsedExpandedNodeID)
+	}
+	return parsedExpandedNodeIDs, nil
+}
+
+func ReadNamespaceArray(endpoint string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
```
kanapuli, line 85: (nit) Formatting seems bit off here.

opcua_plugin/opcua.go
```diff
@@ -67,6 +68,56 @@ func ParseNodeIDs(incomingNodes []string) []*ua.NodeID {
...
+			},
+		},
+	}
+
+	resp, err := client.Read(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	namespaceArray, ok := resp.Results[0].Value.Value().([]string)
```
kanapuli, line 113: (style) This seems like an Hungarian notation in variable names. Please use `namespaces` instead of `namespaceArray`

opcua_plugin/opcua.go
kanapuli: In my opinion , reading the namespace array can be done at the place where we gather Server info while trying to connect.

PR #114, opened 2025-01-24
ENG-2098: add kepware-tests incl. underlying plc's

led0nk:
### Description:
- should add kepware to testing infrastructure
- also should get included in Ci
- should be able to test plc's connected as opc ua-client in kepware
    - via verifying Namespace from NamespaceArray (includes siemens and wago)
    - via checking against an set up Int and Bool (right now only available via siemens)

--- 
### Maybe:
- depending on customer information we may add some more test-cases here (e.g. tag-count/plc-devices connected as clients...)
- later on maybe just try to use some docker containers as opc-ua clients (if available) to increase client count

## Summary by CodeRabbit

- **New Features**
	- Added GitHub Actions workflow for testing the Kepware PLC application.
	- Introduced a comprehensive OPC UA plugin test suite for KepServer EX6.

- **Tests**
	- Implemented detailed test cases for OPC UA connection and data reading.
	- Added test scenarios covering connection scenarios, PLC namespace availability, and value verification.

### DIFF
--- a/.github/workflows/kepware-plc.yml
+++ b/.github/workflows/kepware-plc.yml
@@ -1,0 +1,83 @@
+# Copyright 2023 UMH Systems GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: kepware-plc
+
+on:
+  push:
+    branches:
+      - '**'
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  GO_VERSION: '1.23.0'
+
+concurrency:
+  group: kepware-plc-test
+  cancel-in-progress: true
+
+jobs:
+  go-test-opcua-plc:
+    runs-on:
+      group: arc-runners-tests
+    permissions:
+      packages: write
+      contents: read
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Go
+        uses: ./.github/actions/setup-go
+        with:
+          go_version: ${{ env.GO_VERSION }}
+      - name: Install Ginkgo
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.19.0
+      - name: Install Tcping
+        run: go install github.com/cloverstd/tcping@v0.1.1
+
+      - name: Check Kepware port availability
+        id: check_kepware
+        run: |
+          set +x
+          CURRENT_MINUTE=$(date +%M)
+          CURRENT_HOUR=$(date +%H)
+
+          # make timecheck since we restart kepware-runtime every 2 hours and
+          # don't want to unneccessaryly fail here
+          if [ "$CURRENT_MINUTE" -eq "00" ] && [ $((10#$CURRENT_HOUR % 2)) -eq 0 ]; then
+            sleep 60
+          fi
+
+          URI="${{ secrets.TEST_KEPWARE_ENDPOINT }}"
+          ENDPOINT="${URI#opc.tcp://}"
+          if tcping -c 4 -T 1s "$ENDPOINT" | grep -qi "Connected"; then
+            echo "available=true" >> "$GITHUB_OUTPUT"
+            echo "TEST_KEPWARE_ENDPOINT=${{ secrets.TEST_KEPWARE_ENDPOINT }}" >> "$GITHUB_ENV"
+            echo "using kepware for testing"
+          else
+            echo "available=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Test
+        env:
+          TEST_KEPWARE_USERNAME: ${{ secrets.TEST_KEPWARE_USERNAME }}
+          TEST_KEPWARE_PASSWORD: ${{ secrets.TEST_KEPWARE_PASSWORD }}
+        run: |
+          if [ -z "$TEST_KEPWARE_ENDPOINT" ]; then
+            echo "no kepware-endpoint available for testing"
+            exit 1
+          fi
+          make test
--- a/.github/workflows/kepware-plc.yml
+++ b/.github/workflows/kepware-plc.yml
@@ -22,7 +22,7 @@ on:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
-  GO_VERSION: '1.23.0'
+  GO_VERSION: '1.23'
 
 concurrency:
   group: kepware-plc-test
--- a/opcua_plugin/opcua_kepware_test.go
+++ b/opcua_plugin/opcua_kepware_test.go
@@ -15,12 +15,14 @@ import (
 )
 
 var _ = Describe("Test against KepServer EX6", func() {
-	var endpoint string
-	var username string
-	var password string
-	var input *OPCUAInput
-	var ctx context.Context
-	var cancel context.CancelFunc
+	var (
+		endpoint string
+		username string
+		password string
+		input    *OPCUAInput
+		ctx      context.Context
+		cancel   context.CancelFunc
+	)
 
 	BeforeEach(func() {
 		endpoint = os.Getenv("TEST_KEPWARE_ENDPOINT")
@@ -46,7 +48,13 @@ var _ = Describe("Test against KepServer EX6", func() {
 		}
 	})
 
-	DescribeTable("Connect", func(opcInput *OPCUAInput, errorExpected bool) {
+	DescribeTable("Connect and Read", func(opcInput *OPCUAInput, errorExpected bool, expectedValue any, dynamic bool) {
+		var (
+			messageBatch     service.MessageBatch
+			messageBatch2    service.MessageBatch
+			storedMessage    any
+			assignableNumber json.Number = "10.0"
+		)
 
 		input = opcInput
 		input.Endpoint = endpoint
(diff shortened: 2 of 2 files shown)


opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,324 @@
...
+		err := input.Connect(ctx)
+		if errorExpected {
+			Expect(err).To(HaveOccurred())
+			return
+		}
+		Expect(err).NotTo(HaveOccurred())
+
+		var messageBatch service.MessageBatch
+
+		switch len(input.NodeIDs) {
```
JeremyTheocharis, line 63: How is this swtich statement supposed to work? if you want to check for different type of nodeids, then it would probably make sense to restructure it. or at least comment it explaining what you want to do here and why there is a switch statement on the length of input nodeids. Maybe @kanapuli has a better idea on how to achieve this

JeremyTheocharis: here from chatgpt

Thanks for putting this test code together! It’s overall a solid start, but I noticed a couple of things that could help make it more straightforward:

## 1.	Eliminate (or Clarify) the switch len(input.NodeIDs)
Currently, you’re using a switch on the length of input.NodeIDs to change how you read messages (single node vs. multiple nodes). If we only ever care about two scenarios—1 node vs. 2 nodes—then using a switch might be overkill and can obscure the real goal of each test. A few alternatives:
- Separate Test Cases: Instead of using a single test function with a switch, you could create two distinct DescribeTable entries (one for single-node tests, one for multi-node tests). This would remove the conditional in the middle of your test logic and make it more obvious what’s being tested.
- Comment Clearly: If you really want to keep the switch, add a short comment describing precisely why you’re switching on the number of node IDs, especially in the context of subscriptions vs. direct reads.

## 2.	Leverage Ginkgo’s Table Testing More
The Ginkgo DescribeTable approach allows you to pass in the necessary variations as parameters. Instead of having a single table with a big switch statement in the middle, you could do something like:

```
DescribeTable("Reading single or multiple NodeIDs", 
    func(nodeIDs []string, expectedCount int, dynamic bool) {
      // test logic
    },
    Entry("single node ID", []string{"ns=2;s=Tests.TestDevice.testChangingData"}, 1, false),
    Entry("two node IDs", []string{
        "ns=2;s=Tests.TestDevice.testChangingData", 
        "ns=2;s=Tests.TestDevice.testConstData"
    }, 2, true),
)
```

This way, you tailor each test to either read or subscribe to a single node or multiple nodes, and it’s clearer what the test is doing.

## 3.	Add Explanatory Comments for Subscribe vs. Read
If you’re combining “read once” and “subscribe” logic in the same table, it helps to put a brief comment that clarifies:
- Which part is actually testing “single read” behavior?
- Which part is testing “subscribe to changes”?
This makes it easier for readers (or your future self) to understand each branch.
## 4.	General Style Suggestions
- Keep your test logic as concise as possible—make sure each test focuses on a single aspect (e.g., correct data type, correct subscription behavior).
- When you do a second read (messageBatch2), explicitly state why it’s needed. If you’re validating that data changes over time, mention that clearly in your code comments.

Overall, refactoring the switch statement into separate table entries (or at least adding more clarity to why the switch is there) would make these tests more readable and maintainable. Let me know if you have any questions!

Hope that helps you clean up the tests a bit! Feel free to reach out if you have more questions.

JeremyTheocharis (changes requested):

opcua_plugin/opcua_kepware_test.go
led0nk: This statemant was supposed to work due to the different nodeID-lenght of the tableentries, but this was just a unclear workaround. It wasn't supposed to differentiate between different nodeID-types.

I refactored the test to have a clearer approach and idea of how it should work.
So we now just differentiate if it's a dynamic or non-dynamic value and only if we'll check that there are data changes.

We could also move both tables together, but i wanted to differ between the kepware-values (only set in kepserver) and the one's that are coming truely via plc as an opc-client into kepware.

kanapuli (approved):

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,314 @@
...
+	)
+
+	BeforeEach(func() {
+		endpoint = os.Getenv("TEST_KEPWARE_ENDPOINT")
+		username = os.Getenv("TEST_KEPWARE_USERNAME")
+		password = os.Getenv("TEST_KEPWARE_PASSWORD")
+
+		if endpoint == "" || username == "" || password == "" {
+			Skip("Skipping test: environmental variables are not set")
+			return
```
kanapuli, line 34: You don't need a return statement here. Ginkgo will take care of skipping this suite automatically when the condition passes.
Having a `return` statement fails the test @led0nk

kanapuli, line 34:
```suggestion
```

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,314 @@
...
```
kanapuli, line 172:
```suggestion
```

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,314 @@
...
+			err := input.Client.Close(ctx)
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		if cancel != nil {
+			cancel()
+		}
+	})
+
+	DescribeTable("Test if PLC-Namespaces are available", func(namespace string, nodeID *ua.NodeID, contains bool) {
```
kanapuli, line 189: (readability) the variable name `contains` does not self explain its purpose. Can we have a better variable naming please ?

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,314 @@
...
+		),
+		Entry(
+			"should fail due to incorrect namespace",
+			"totally wrong namespace",
+			ua.NewStringNodeID(2, "Wago.play.Server.NamespaceArray"),
+			false,
+		),
+	)
+
+	DescribeTable("check for correct values", func(opcInput *OPCUAInput, expectedValue any, dynamic bool) {
```
kanapuli, line 248: (readability) Same for the variable name `dynamic`. It is not clear what `dynamic` means here. May be a better variable name should help.

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,314 @@
...
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(message).To(BeAssignableToTypeOf(assignableNumber))
+				Expect(message).NotTo(Equal(storedMessage))
+			}
+		}
+	},
+		Entry("should check if message-value is true", &OPCUAInput{
+			Username:                   "",
+			Password:                   "",
```
kanapuli, line 301: (nit) Not necessary to explicitly set these empty values here. Go will initialize the struct with empty values.

opcua_plugin/opcua_kepware_test.go
kanapuli: image

led0nk: Since this appears in many other tests, should we also remove it from there? @kanapuli 
`return` doesn't seem to fail the test, i mean it doesn't make sense here but as far as i understood it just doesn't get called, correct?

kanapuli: No we don't need to remove in other places. We can refactor them while touching those code to keep the changes smaller here

kanapuli (approved):

opcua_plugin/opcua_kepware_test.go
led0nk: maybe we should keep in mind then, that we could refactor many test-cases to recommended table-driven tests, since especially the "connecting"-tests are mostly equal

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,308 @@
...
+			NodeIDs:                    ParseNodeIDs([]string{"ns=2;s=SiemensPLC_main.main.ServerInterfaces.Server _interface_1.test"}),
+			AutoReconnect:              true,
+			ReconnectIntervalInSeconds: 5,
+		}, true, false),
+		Entry("should return data changes on subscribe", &OPCUAInput{
+			NodeIDs:          ParseNodeIDs([]string{"ns=2;s=SiemensPLC_main.main.ServerInterfaces.Server _interface_1.counter"}),
+			SubscribeEnabled: true,
+		}, nil, true),
+	)
+
```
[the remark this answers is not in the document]
led0nk: implemented, but not exaclty like this but with duplicated code

JeremyTheocharis (approved):

opcua_plugin/opcua_kepware_test.go
```diff
@@ -0,0 +1,279 @@
...
+	AfterEach(func() {
+		if input.Client != nil {
+			err := input.Client.Close(ctx)
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		if cancel != nil {
+			cancel()
+		}
+	})
```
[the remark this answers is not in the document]
JeremyTheocharis: good point @led0nk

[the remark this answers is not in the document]
led0nk: I already talked with @kanapuli about this, since this happens in some other test-suites,too. Fixed it here, will somewhen fix it in other test-suites.

.github/workflows/kepware-plc.yml
```diff
@@ -0,0 +1,83 @@
...
+name: kepware-plc
+
+on:
+  push:
+    branches:
+      - '**'
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  GO_VERSION: '1.23.0'
```
Scarjit, line 25: This could be 1.23.5, but not blocking this pr.

Scarjit (approved):

.github/workflows/kepware-plc.yml
led0nk: but does that make sense, since we're using `1.23` in benthos-umh?

led0nk: nvm

PR #118, opened 2025-01-30
ENG-2389: add trusted server option via fingerprint

led0nk:
### Description:
- adds the option to set the server as a "trusted" server by providing the fingerprint
- the fingerprint is the sha1 sum of the server-certificate
- the option is only set when using the flag `fingerprint: <servers-fingerprint>` in input.yaml
- adds testing with wago and siemens-plcs
- adds tests to workflow

The fingerprint will get checked while setting up a connection. The filtering happens while fetching the endpoints since we don't have to worry about implementing it into `connectWithSecurity` as well as `attemptBestEndpointConnection`. Thererfore it will always be used if the user set's the `fingerprint`.


README.md
```diff
@@ -125,6 +125,7 @@ input:
     insecure: false | true # DEPRECATED, see below
     securityMode: None | Sign | SignAndEncrypt # optional (default: unset)
     securityPolicy: None | Basic256Sha256  # optional (default: unset)
+    fingerprint: 'sha1-fingerprint-of-cert' # optional (default: unset)
```
Scarjit, line 128: Please use a secure hash like sha256, sha512 or sha3 instead of sha1.
See also: https://en.wikipedia.org/wiki/SHA-1#Attacks

opcua_plugin/connect.go
```diff
@@ -172,8 +174,80 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+		// parse the DER-format certificate
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			g.Log.Warnf("Invalid server certificate for endpoint %s. Skipping...", ep.EndpointURL)
+			continue
+		}
+
+		// calculating the checksum of the certificate (sha1 is needed here)
+		// and convert the array into a slice for encoding
+		shaFingerprint := sha1.Sum(cert.Raw)
```
Scarjit, line 234: See my first comment (sha1 usage)

Scarjit (changes requested):

opcua_plugin/opcua_plc_test.go
```diff
@@ -250,6 +288,42 @@ var _ = Describe("Test Against WAGO PLC", Serial, func() {
...
 	})
 
+	Describe("Connect to trusted server", func() {
+		It("should connect successfully", func() {
+			fingerprint = os.Getenv("TEST_WAGO_FINGERPRINT")
+
+			// we only want to skip the fingerprint test here
+			if fingerprint == "" {
+				Skip("Skipping test: environment variable not set")
+				return
```
kanapuli, line 298:
```suggestion
```

opcua_plugin/opcua_plc_test.go
```diff
@@ -250,6 +288,42 @@ var _ = Describe("Test Against WAGO PLC", Serial, func() {
...
+			input = &OPCUAInput{
+				Endpoint:    endpoint,
+				Username:    "",
+				Password:    "",
+				NodeIDs:     nil,
+				Fingerprint: "test123", // incorrect certificate fingerprint
+			}
+
+			err := input.Connect(ctx)
+			Expect(err).To(HaveOccurred())
```
JeremyTheocharis, line 323: see comment above

opcua_plugin/connect.go
```diff
@@ -172,8 +174,80 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+		return endpoints, nil
+	}
+
+	var filteredEP []*ua.EndpointDescription
+
+	for _, ep := range endpoints {
+		// if the endpoint doesn't provide a server certificate, we skip it
+		// we could potentially return here since this should be the same for all endpoints
+		if len(ep.ServerCertificate) == 0 {
+			g.Log.Warnf("Endpoint %s doesn't provide any server certificate. Skipping...", ep.EndpointURL)
```
JeremyTheocharis, line 208: see also other comment about warnings

opcua_plugin/connect.go
```diff
@@ -172,8 +174,80 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+// filterEndpointsByFingerprint will filter the endpoints provided by the server if the fingerprint is set
+//
+// This function will check on each element of the fetched endpoints, if the fingerprint is correct.
+// Therefore we can check if the server is "trusted" by the client. It should check on the server certificate
+// of each endpoint if the fingerprint (sha1) matches with the fingerprint from input.yaml.
+//
+// **Why This Function is Needed:**
+//   - To ensure the client connects to the correct server
+func (g *OPCUAInput) filterEndpointsByFingerprint(endpoints []*ua.EndpointDescription) ([]*ua.EndpointDescription, error) {
+	if g.Fingerprint == "" {
```
JeremyTheocharis, line 197: what about the case that we don;t connect encrypted? what of the cases where the user selects their own security info (e.g., by setting security policgy)?

JeremyTheocharis, line 197: what about the case that we don;t connect encrypted? what of the cases where the user selects their own security info (e.g., by setting security policgy)?

README.md
```diff
@@ -187,6 +188,19 @@ input:
     securityPolicy: Basic256Sha256
 ```
 
+##### Fingerprint
+
+You can provide the fingerprint of your OPC-UA-Servers certificate, which should be a sha1-hash.
```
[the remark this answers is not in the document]
JeremyTheocharis: Suggestion:

```
##### Server Certificate Fingerprint
**Key**: `serverCertificateFingerprint`  
**Description**:  
Use this field to explicitly trust the server’s certificate. When specified, only endpoints matching this fingerprint will be accepted. This ensures the client connects to the correct server and helps prevent man-in-the-middle attacks.

> **Important**  
> - **If you omit `serverCertificateFingerprint`,** the client will still attempt to connect.  
>   - Initially, it will log an **info message** to remind you to set `serverCertificateFingerprint`.  
>   - Future releases may escalate this to a **warning** that blocks deployment in certain environments.  
```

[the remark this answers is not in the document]
JeremyTheocharis: If your server’s certificate changes (e.g., renewal, new server), update the serverCertificateFingerprint accordingly. Otherwise, the plugin will reject the connection, signaling a potential security issue or misconfiguration.

opcua_plugin/connect.go
```diff
@@ -172,8 +174,80 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+	if err != nil {
+		g.Log.Errorf("Failed to filter endpoint by fingerprint: %s", err)
+		return nil, err
+	}
+
 	// For multiple endpoints, adjust their hosts as per user specification.
-	return adjustedEndpoints, nil
+	return filteredEndpoints, nil
+}
+
```
JeremyTheocharis, line 187: during initial connect (without fingerprint ser):

### 1. When the `serverCertificateFingerprint` is Empty

#### Step 1: Log an *Info* Message
*(Initial approach, does not block deployment)*

```go
// Info-level log when user did not specify `serverCertificateFingerprint`
g.Log.Infof(
  "[OPC UA Plugin] No 'serverCertificateFingerprint' was provided. " +
  "We strongly recommend specifying 'serverCertificateFingerprint=xxxx' to verify the server's identity " +
  "and avoid potential security risks. Future releases may escalate this to a warning that prevents deployment."
)

Comment (JeremyTheocharis) [reply opcua_plugin/connect.go:187]: // Warning-level log in future release
g.Log.Warnf(
  "[OPC UA Plugin] Missing 'serverCertificateFingerprint'. " +
  "Connections without a trusted certificate fingerprint can be insecure. " +
  "Please set 'serverCertificateFingerprint=xxxx' to ensure a trusted connection. " +
  "Continuing without a fingerprint may be blocked in future versions."
)

File: opcua_plugin/connect.go [hunk 9]
```diff
@@ -172,8 +174,80 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+			g.Log.Warnf("Fingerprint of endpoint %s doesn't match, expected: '%s', got: '%s'. Skipping...", ep.EndpointURL, g.Fingerprint, epFingerprint)
+			continue
+		}
+
+		filteredEP = append(filteredEP, ep)
+
+		// return an error if there are no endpoints with matching fingerprints - to see failure
+	}
+	if len(filteredEP) == 0 {
+		return nil, fmt.Errorf("no endpoints with matching certificate fingerprint found: '%s' ", g.Fingerprint)
```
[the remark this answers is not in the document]
JeremyTheocharis: g.Log.Warnf(
  "Endpoint %s certificate fingerprint mismatch. Expected: '%s', got: '%s'. "+
  "Either the server’s certificate was intentionally updated, or you are connecting to the wrong server. "+
  "If intentional, please set the new 'serverCertificateFingerprint' in your configuration. "+
  "Otherwise, double-check your security settings.",
  ep.EndpointURL, g.ServerCertificateFingerprint, epFingerprint,
)

opcua_plugin/connect.go
JeremyTheocharis: // Warning-level log in future release
g.Log.Warnf(
  "[OPC UA Plugin] Missing 'serverCertificateFingerprint'. " +
  "Connections without a trusted certificate fingerprint can be insecure. " +
  "Please set 'serverCertificateFingerprint=xxxx' to ensure a trusted connection. " +
  "Continuing without a fingerprint may be blocked in future versions."
)

JeremyTheocharis: // Warning-level log in future release
g.Log.Warnf(
  "[OPC UA Plugin] Missing 'serverCertificateFingerprint'. " +
  "Connections without a trusted certificate fingerprint can be insecure. " +
  "Please set 'serverCertificateFingerprint=xxxx' to ensure a trusted connection. " +
  "Continuing without a fingerprint may be blocked in future versions."
)

README.md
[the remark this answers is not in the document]
JeremyTheocharis: As an outlook (maybe create a linear issue for this): the companion during browsing could fetch this fignerprint and set it directly in the frontend by default, so the user would never need to set this themselves (except when the server would change). @kanapuli

JeremyTheocharis (changes requested):

README.md
led0nk: switched to sha3

opcua_plugin/connect.go
led0nk: (same as above)

README.md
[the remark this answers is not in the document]
led0nk: I kind of took your suggestions and yeah it definitely makes sense to explain this a bit more in detail.

opcua_plugin/opcua_plc_test.go
```diff
@@ -154,6 +155,42 @@ var _ = Describe("Test Against Siemens S7", Serial, func() {
...
```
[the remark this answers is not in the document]
led0nk: updated

opcua_plugin/opcua_plc_test.go
led0nk: as above

opcua_plugin/connect.go
led0nk: Well this get's always called but returns early if no fingerprint provided.
We use this when fetching all endpoints, to filter and inform/warn about non-matching fingerprints (if provided) or error out if none are found. We could move the statement `if g.Fingerprint` out to the `FetchAllEndpoints` method to only get called if the user provides any fingerprints, but i'm unsure if it makes sense here.


So this works for:
- no encryption at all
- selecting their own security info
- direct connect

README.md
```diff
@@ -187,6 +188,19 @@ input:
     securityPolicy: Basic256Sha256
 ```
 
+##### Fingerprint
```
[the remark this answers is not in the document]
led0nk: done

opcua_plugin/connect.go
led0nk: So i refactored this a bit, since i looked up the `opcua.GetEndpoints` and it didn't make sense before to check on each endpoint and filter the endpoints. So what we do now is:
1. check if fingerprint is set
2. if fingerprint isn't set -> provide some info message that this is potentially unsafe and may lead to a failure of the deployment in future releases
3. if fingerprint is set -> check from the first endpoint (since the first one should always be available, or we will fail anyways) 
4. if fingerprint matches continue with connecting attempt
5. if fingerprint doesn't match provide some Information regarding the mismatch (including the server's fingerprint -> so the user could copy/paste the fingerprint (sha3)

opcua_plugin/connect.go
led0nk: refactored

opcua_plugin/connect.go
```diff
@@ -172,8 +174,80 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+		}
+
+		// calculating the checksum of the certificate (sha1 is needed here)
+		// and convert the array into a slice for encoding
+		shaFingerprint := sha1.Sum(cert.Raw)
+		epFingerprint := hex.EncodeToString(shaFingerprint[:])
+
+		// to have some information on the mismatched fingerprints
+		if epFingerprint != g.Fingerprint {
+			g.Log.Warnf("Fingerprint of endpoint %s doesn't match, expected: '%s', got: '%s'. Skipping...", ep.EndpointURL, g.Fingerprint, epFingerprint)
```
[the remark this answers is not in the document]
led0nk: also refactored

opcua_plugin/connect.go
```diff
@@ -502,6 +563,28 @@ func (g *OPCUAInput) connect(ctx context.Context) error {
...
+		err = g.checkServerCertificateFingerprint(endpoints[0])
+		if err != nil {
+			g.Log.Infof("Error while trying to match the server's certificate fingerprint: %s", err)
+			return err
+		}
+	default:
+		// may be escalated to Warning-Level in future releases
+		g.Log.Infof(
+			"No 'serverCertificateFingerprint' was provided. " +
+				"We strongly recommend specifying 'serverCertificateFingerprint=xxxx' to verify the server's identity " +
```
JeremyTheocharis, line 583: this here should then print the actual serverCertificateFingerprint, to make it easier for the user to copy it

opcua_plugin/connect.go
```diff
@@ -176,6 +179,64 @@ func (g *OPCUAInput) FetchAllEndpoints(ctx context.Context) ([]*ua.EndpointDescr
...
+// fingerprint is correct. Therefore we can check if the server is "trusted"
+// by the client. It should check on the server certificate of the endpoint
+// if the fingerprint (sha3) matches with the fingerprint from input.yaml.
+//
+// **Why This Function is Needed:**
+//   - To ensure the client connects to the correct server
+func (g *OPCUAInput) checkServerCertificateFingerprint(endpoint *ua.EndpointDescription) error {
+
+	// If the endpoint doesn't provide a server certificate, we return here
+	// This is not expected to happen, since an OPC-UA server should always
```
JeremyTheocharis, line 194: it should only have a certificate when using encrypted connection. probably makes sense to include a check here, that once a fingerprint is selected, we only accept encrypted connections

JeremyTheocharis: with this we should then be able to prevent a downgrade attack, that accidently benthos selects an unencrypted certificate even though the fingerprint is set

JeremyTheocharis (changes requested):

led0nk:
image

opcua_plugin/connect.go
```diff
@@ -419,27 +422,33 @@ func (g *OPCUAInput) logCertificateInfo(certBytes []byte) {
...
 	orderedEndpoints := g.orderEndpoints(endpoints, authType)
 	for _, currentEndpoint := range orderedEndpoints {
+
+		// If the endpoints' 'securityMode' is 'SignAndEncrypt' & the endpoints'
+		// 'securityPolicy' is either Basic256Sha256, Basic256 or Basic128Rsa15
+		// print an information to the user that he should set the needed flags.
+		if currentEndpoint.SecurityMode == ua.MessageSecurityModeSignAndEncrypt &&
+			!isNoSecurityEndpoint(currentEndpoint) {
+			serverCertFingerprint, _ := g.getServerCertificateFingerprint(currentEndpoint)
+			g.Log.Infof("Encrypted endpoint was found, please provide the following "+
```
JeremyTheocharis, line 441: Update the error message to show the exact securityPolicy that the user should select. Also include why the user should provide this information (to enable encryption)

JeremyTheocharis (changes requested):

opcua_plugin/connect.go
```diff
@@ -538,6 +567,36 @@ func (g *OPCUAInput) connectWithSecurity(ctx context.Context, endpoints []*ua.En
...
+func (g *OPCUAInput) unencryptedConnect(
+	ctx context.Context,
+	endpoints []*ua.EndpointDescription,
+	authType ua.UserTokenType,
+) (*opcua.Client, error) {
+
+	//TODO: specify
+	g.Log.Infof("Establishing an unencrypted connection...")
+
+	if g.DirectConnect || len(endpoints) == 0 {
```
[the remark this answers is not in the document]
led0nk: No it was like this:
 1.  if the user provided the DirectConnect flag or if we didn't get any endpoints
 2. otherwise `attemptBestEndpointConnection`

https://github.com/united-manufacturing-hub/benthos-umh/blob/4db372894e0697bf1d8ebfd80358197f0111ae4b/opcua_plugin/connect.go#L486-L560

led0nk:
ignore those commits for now, i just started resolving

opcua_plugin/connect.go
```diff
@@ -158,10 +164,18 @@ func (g *OPCUAInput) checkForSecurityEndpoints(
...
 
+				// we use the openConnection func here to pre-create the cert and with
+				// the connect-attempt we send it over to the OPC-UA server so the user,
+				// can proceed to trust this certificate and only has to copy the snippet
+				// below
+				c, err := g.openConnection(ctx, ep, authType)
+				if err == nil {
+					c.Close(ctx)
+				}
+
```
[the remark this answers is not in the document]
led0nk: that's exactly what i was about to do, but gonna resolve the conflicts first

led0nk:
@coderabbitai pause

led0nk:
@coderabbitai resume

led0nk:
image


So now we have 2 different starting points here:
1) without any specification of security
    - this will check for `security-endpoints` and provide you with a snippet + send your certificate to the server
    - then you could just go to the servers-endpoint settings and trust the certificate
    - then you'll have to paste this snippet into your input.yaml
    - otherwise IF your server has an endpoint for `None` (securityPolicy) it will connect afterwards
 2) you specify `securityMode` and `securityPolicy`
    - as usual you get your certificate
    - depending on the opc-ua server it will either auto-connect or you have to manually trust the cert
    - if you provided a `serverCertificateFingerprint` it will check on that and connect if it matches
    - if you didn't provide a `serverCertificateFingerprint` it will inform you to set that to `xyz` but continue with connection

Since the "encrypted"-connection doesn't work well with Siemens right now, I just prepared tests / workflow for the S7-Fingerprint, but the test isn't included here, so it should get skipped.

opcua_plugin/authentication.go
```diff
@@ -138,3 +138,55 @@ func (g *OPCUAInput) getEndpointIfExists(
...
+		securityPolicyOK = false
+	default:
+		securityPolicyOK = false
+	}
+
+	if securityModeOK && securityPolicyOK {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
led0nk: while this would make sense here, i think it's just more "readable" in a switch-statement here

Scarjit (approved):

README.md
```diff
@@ -169,23 +171,38 @@ input:
 
 ##### Security Options
 
-Security Mode: This defines the level of security applied to the messages. The options are:
-- None: No security is applied; messages are neither signed nor encrypted.
-- Sign: Messages are signed for integrity and authenticity but not encrypted.
-- SignAndEncrypt: Provides the highest security level where messages are both signed and encrypted.
+OPC UA supports various security modes and security policies. These options define how messages are signed or encrypted and which cryptographic algorithms are used. In the configuration, you can specify the following:
```
JeremyTheocharis, line 174: maybe state that in order to securely connect, all the following parameters need to be set ("strict mode")

JeremyTheocharis: also state, that security doesn't need to be set necessarily, benthos-umh is able to auto-negotiate

README.md
```diff
@@ -169,23 +171,38 @@ input:
...
+  - If you specify this field, the client will verify that the server’s certificate matches the given fingerprint. If there’s a mismatch, the connection is rejected.
+  - If omitted while **still using encryption** (`Sign` or `SignAndEncrypt`), the client will attempt to connect and then **log** the server’s actual fingerprint. You can copy that fingerprint into your config to be certain you’re connecting to the intended server.
+  - In future releases, omitting the fingerprint may become a warning or block deployment in certain environments.
 
+- **Client Certificate**:
+  - **Key**: `clientCertificate`
+  - **Description**: A Base64‐encoded PEM bundle (certificate + private key).
+  - When using encryption (`Sign` or `SignAndEncrypt`), the client must present a certificate to the server. If you **do not** provide one, the system **auto‐generates** a random certificate at startup.
+  - The auto‐generated certificate is logged in Base64 so you can copy/paste it into your configuration. This allows the server to trust the same client certificate across restarts instead of generating a new one each time.
```
JeremyTheocharis, line 202: mention at least here, if not in the overview how it works with the server needing to trust the client certificate

.github/workflows/opcua-plc.yml
```diff
@@ -82,6 +84,7 @@ jobs:
           if tcping -c 4 -T 1s "$ENDPOINT" | grep -qi "Connected"; then
             echo "available=true" >> "$GITHUB_OUTPUT"
             echo "TEST_WAGO_ENDPOINT_URI=${{ secrets.TEST_WAGO_ENDPOINT_URI }}" >> "$GITHUB_ENV"
+            echo "TEST_WAGO_FINGERPRINT=${{ secrets.TEST_WAGO_FINGERPRINT }}" >> "$GITHUB_ENV"
```
JeremyTheocharis, line 87: why is the wago fingerprint as a secret, and the kepware one in the test there not? either dont go for it as a secret at all, or if yes, do it the same for all of them

JeremyTheocharis (changes requested):
some small stuff, otherwise looks great :)

.github/workflows/opcua-plc.yml
led0nk: i guess this just slipped through, because earlier on i didn't want to set that as secrets since certificates can change over time and kepware doesn't have a fallback here, but changed that

README.md
```diff
@@ -171,10 +171,12 @@ input:
 
 ##### Security Options
 
+> **Note**: In order to ensure a fully secure connection, we strongly recommend specifying **all** of the following parameters. However, if you omit them, **benthos-umh** will attempt to discover and prepare the most secure endpoint automatically, and it will also log the recommended settings (security mode/policy, server fingerprint, and generated client certificate) so that you can easily configure them once you are ready
```
JeremyTheocharis, line 174: let's adjust it to be precise: 

> To ensure a fully secure connection, you must explicitly configure all of the following security options. However, if these settings seem overwhelming, you can leave them unspecified. In that case, **benthos-umh** will automatically scan for and connect to available endpoints until it succeeds—and then it will log the recommended security settings for your future configuration.

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #123, opened 2025-02-04
ENG-2386: implement "encryption" certificates for "encrypted" connection

led0nk:
## Description:

What we are trying to achieve here:
 - implementation of `Basic256Sha256`
 - implementation of `Basic256` (depracated)
 - implementation of `Basic128Rsa15` (depracated)
 
User flow:
 - set an input.yaml e.g.:
```yaml
input:
  opcua:
    endpoint: <your-endpoint>
    username: ''
    password: ''
    subscribeEnabled: true
    useHeartbeat: true
    securityMode: SignAndEncrypt
    securityPolicy: Basic256Sha256
    nodeIDs:
      - i=2255

```
**important: set the fields `securityMode` and `securityPolicy`**

- benthos-umh logs will then show a warning, since the connection is not successfull but we also print out an info-message to tell the user he has to trust the client on the server-side
- when he trusted the client's certificate (which is specified in the info-message via `application-uri`,  benthos-umh will retry connection and then get the specified nodes / browse first

The easiest way for user-experience would be to set the option `Trust all clients` in the OPC-UA-Servers configuration, which depends on the Software you're running on.

### Basic256Sha256
Basic256Sha256-kepware
**See the logs of this run here:** 
[basic256sha256.log](https://github.com/user-attachments/files/18658231/basic256sha256.log)

### Basic256
Basic256-kepware
**See the logs of this run here:**
[basic256.log](https://github.com/user-attachments/files/18658230/basic256.log)

### Basic128Rsa15
Basic128Rsa15-kepware
[basic128rsa15.log](https://github.com/user-attachments/files/18658229/basic128rsa15.log)

## What needs to be done here

 - [x] testing with Siemens-OPC-Server failed due to Siemens-settings (doesn't even work with UAExpert)
 - [x] testing with Prosys-OPC-Server failed due to Prosys-settings (doesn't even work with UAExpert) 
 - [x] rewrite ginkgo-tests for encryption

As an outlook:
 - [ ] maybe silently try to connect / or with a limited amount of times without erroring out (maybe within every 5 seconds) so that we don't instantly throw out an error here:
https://github.com/united-manufacturing-hub/benthos-umh/blob/4db372894e0697bf1d8ebfd80358197f0111ae4b/opcua_plugin/connect.go#L394-L426

-> this would lead to benthos retrying to connect every 5 seconds, which makes it much harder for the user to get the correct `application-uri` from the logs (as you can see in all of the provided logs)
 

## Summary by CodeRabbit

- **Documentation**
  - Revised configuration instructions for secure connections.
  - Removed deprecated insecure mode and expanded available security policy options.
  - Introduced an option for specifying a fixed Base64-encoded client certificate.

- **Chores**
  - Upgraded several dependencies to improve compatibility and performance.

- **New Features**
  - Enhanced certificate handling with more robust parsing, flexible certificate generation, and endpoint security categorization.
  - Added support for multiple security policies and modes in OPC UA connections.
  - Introduced a new field for accepting a Base64-encoded client certificate in the configuration.

- **Tests**
  - Expanded test coverage to validate a broad range of security configurations.
  - Streamlined test cases for selecting a custom security policy using a table-driven approach.


opcua_plugin/connect.go
```diff
@@ -37,31 +37,53 @@ func (g *OPCUAInput) GetOPCUAClientOptions(selectedEndpoint *ua.EndpointDescript
...
-	// Generate certificates if Basic256Sha256
-	if selectedEndpoint.SecurityPolicyURI == ua.SecurityPolicyURIBasic256Sha256 {
-		randomStr := randomString(8) // Generates an 8-character random string
-		clientName := "urn:benthos-umh:client-" + randomStr
-		certPEM, keyPEM, err := GenerateCert(clientName, 2048, 24*time.Hour*365*10)
-		if err != nil {
-			g.Log.Errorf("Failed to generate certificate: %v", err)
-			return nil, err
-		}
+	// Generate certificates if we don't connect without Security
```
JeremyTheocharis, line 40: maybe no double negation? Generate certificates, when we connect with security

opcua_plugin/connect.go
```diff
@@ -37,31 +37,53 @@ func (g *OPCUAInput) GetOPCUAClientOptions(selectedEndpoint *ua.EndpointDescript
...
+
+			// Convert PEM to X509 Certificate and RSA PrivateKey for in-memory use.
+			cert, err := tls.X509KeyPair(certPEM, keyPEM)
+			if err != nil {
+				g.Log.Errorf("Failed to parse certificate: %v", err)
+				return nil, err
+			}
+			// cache the tls.Certificate for future calls
+			g.cachedTLSCertificate = &cert
+			g.Log.Infof("The clients certificate was created, to use an encrypted connection "+
```
JeremyTheocharis, line 74: this only make sense when the certificate seed was choosen by the user / is fixed. if its random, it will not work

opcua_plugin/opcua.go
```diff
@@ -46,7 +47,8 @@ var OPCUAConfigSpec = service.NewConfigSpec().
...
 	Field(service.NewBoolField("autoReconnect").Description("Set to true to automatically reconnect to the OPC UA server when the connection is lost. Defaults to 'false'").Default(false)).
-	Field(service.NewIntField("reconnectIntervalInSeconds").Description("The interval in seconds at which to reconnect to the OPC UA server when the connection is lost. This is only used if `autoReconnect` is set to true. Defaults to 5 seconds.").Default(5))
+	Field(service.NewIntField("reconnectIntervalInSeconds").Description("The interval in seconds at which to reconnect to the OPC UA server when the connection is lost. This is only used if `autoReconnect` is set to true. Defaults to 5 seconds.").Default(5)).
+	Field(service.NewStringField("certificateSeed").Description("The certificate seed is a secret string provided by the user, which will then get hashed to create the clients certificate. This is needed to ensure an explicit trusted client, which is significant for an encrypted connection. If not set, this will be provided by a random created string.").Default(""))
```
JeremyTheocharis, line 51: can we put that into the README as well? Maybe also rename that:

The certificate seed is used to deterministically create the client certificate when using encryption. If not set, then the certificate will be random upon restart, which means the server must trust all client certificates. If the server requires to trust each single client certificate, then the certificate seed must be set to a random (but therefore fixed) string with at least 64 characters.

opcua_plugin/connect.go
```diff
@@ -37,31 +37,53 @@ func (g *OPCUAInput) GetOPCUAClientOptions(selectedEndpoint *ua.EndpointDescript
...
-			g.Log.Errorf("Failed to generate certificate: %v", err)
-			return nil, err
-		}
+	// Generate certificates if we don't connect without Security
+	if selectedEndpoint.SecurityPolicyURI != ua.SecurityPolicyURINone {
+		if g.cachedTLSCertificate == nil {
+			if g.CertificateSeed == "" {
+				// Generate an 8-character random string if no 'certificateSeed'
+				// provided by the user.
+				g.CertificateSeed = randomString(8)
```
JeremyTheocharis, line 46: I would increase the random string from 8 to 64

opcua_plugin/connect.go
JeremyTheocharis: if its random, you can instead output:

> The client certificate was generated randomly upon starting up, and will change upon every restart. In order to fix it, e.g., to trust the client certificate in the server, please set the certificateSeed cofniguration variable to a random 64 length string (e.g., generated by a random generator) or alternatively use this to fix the current certificate: %s), (g.CertificateSeed)

.DS_Store
JeremyTheocharis, line 1: :)

JeremyTheocharis (changes requested):
## Overall Impressions

- **Great Feature Work:**  
  The implementation of encryption certificates for Basic256Sha256, Basic256, and Basic128Rsa15 is a key step for securing OPC UA connections. The work is promising and sets a strong foundation for future enhancements.

- **User Experience Focus:**  
  Although this is a security feature, it’s crucial that even non-security experts can understand and use it. The logs, configuration messages, and documentation should be clear, friendly, and guide the user seamlessly—mirroring the user-centric experience of products like Apple and Linear.

## Detailed Technical Feedback

1. **Certificate Generation & Seed Handling (opcua_plugin/connect.go & generate_cert.go):**
   - **Random String Length:**  
     Increase the default certificate seed random string length from 8 to 64 characters for better entropy.
   - **User Messaging:**  
     Log a clear message that includes a copy-paste-ready configuration snippet when a random certificate seed is generated. For example:  
     _"The client certificate was generated randomly upon startup and will change on every restart. To use the current dynamically generated certificate as a fixed certificate, copy the following configuration snippet into your config: certificateSeed: \"<generated_seed>\"." (where generated_seed is the current auto-generated g.certificateSeed)_  

2. **Configuration & Documentation (opcua_plugin/opcua.go):**
   - **Rewording of the `certificateSeed` Field:**  
     Update the description to:  
     _"The certificate seed is used to deterministically generate the client certificate when using encryption. If not set, a new (random) certificate will be generated at every restart, requiring the OPC UA server to trust all client certificates. For a fixed certificate, set this value to a fixed random string of at least 64 characters."_


Example:
```
if g.CertificateSeed == "" {
            // Generate a 64-character random seed if none is provided
            g.CertificateSeed = randomString(64)
            // Prepare a configuration snippet for the user to copy into their config
            configSnippet := fmt.Sprintf("certificateSeed: \"%s\"", g.CertificateSeed)
            g.Log.Infof("The client certificate was generated randomly upon startup and will change on every restart.\n"+
                "To use a fixed certificate, please copy and paste the following line into your configuration:\n%s", configSnippet)
        }
```


Here an example for the README:

### Encryption Certificates & Certificate Seed

To enhance the security of OPC UA connections, benthos-umh now supports generating client certificates when using encryption. This feature operates in two modes:

1. **Dynamic Mode (Default):**  
   If no certificate seed is provided, benthos-umh automatically generates a new random certificate seed (64 characters long) at each startup. This results in a new client certificate every time the service restarts, meaning the OPC UA server must be configured to trust all client certificates.

2. **Fixed Mode:**  
   For environments where the OPC UA server requires each client certificate to be explicitly trusted, you can set a fixed certificate seed. By providing a fixed random string (of at least 64 characters) via the configuration, the client certificate will be deterministically generated on every startup, ensuring consistency.

#### How It Works

When operating in Dynamic Mode, the application logs a clear message that includes a copy-paste-ready configuration snippet. For example, you might see a log message like this:

> **Client Certificate Notice:**  
> The client certificate was generated randomly upon startup and will change on every restart.  
> To use the current dynamically generated certificate as a fixed certificate, copy the following configuration snippet into your config:  
> `certificateSeed: "REPLACE_WITH_GENERATED_SEED"`

Simply copy that line into your configuration file to lock the certificate to the current seed, ensuring that the same certificate is generated on every restart.

#### Configuration Example

Below is an example of how to set up the OPC UA input with a fixed certificate seed in your `benthos.yaml`:

```yaml
input:
  opcua:
    endpoint: 'opc.tcp://localhost:46010'
    nodeIDs: ['ns=2;s=IoTSensors']
    securityMode: SignAndEncrypt
    securityPolicy: Basic256Sha256
    certificateSeed: "your-fixed-64-character-random-string"
```

Replace `"your-fixed-64-character-random-string"` with a secure, random string that is at least 64 characters long. If you leave out the `certificateSeed` field, benthos-umh will generate one automatically at startup, which means the client certificate will change with every restart.

opcua_plugin/connect.go
led0nk: We will now take the client-name based on which the certificate was created and print it out here.

opcua_plugin/opcua.go
led0nk: updated to `clientCertificateSeed`and added to Readme to "Security"

led0nk:
I am currently having trouble since the tests are failing, which doesn't really make sense to me, because they're running like exactly the same configuration as I run the application. The Wago-Endpoint-URI is correctly set and available, also it should like in the application-run (shown in the log-file) automatically accept the certificate.

image

```yaml
input:
  opcua:
    endpoint: opc.tcp://10.13.37.54:4840
    username: ''
    password: ''
    securityPolicy: Basic256Sha256
    securityMode: SignAndEncrypt
    nodeIDs:
      - i=84

```

[wago-basic256sha256.log](https://github.com/user-attachments/files/18688988/wago-basic256sha256.log)

led0nk:
### User flow (updated):
#### 1.)
You set up the input.yaml with the flags `securityMode: SignAndEncrypt | Sign` & `securityPolicy: Basic256Sha256 | Basic 256 | Basic128Rsa15`. Like this:
```yaml
---
input:
  opcua:
    endpoint: opc.tcp://ip-address:port
    username: ''
    password: ''
    subscribeEnabled: true
    useHeartbeat: true
    securityMode: SignAndEncrypt
    securityPolicy: Basic256Sha256
    nodeIDs:
      - i=2255
```

Then you'll get an Info-message which tells you to go over to the OPC-UA-Server an trust the client's certificate which should be named `benthos-umh-predefined. On top of that you get an Info-message which tells you to set the `clientCertificateSeed` to a certain string (as shown below and in logs). 
So you can go down 2 ways from here:
Either first go to the server and trust the certificate and afterwards paste the provided `clientCertificateSeed`.
Or paste the `clientCertificateSeed` rerun the application and trust the certificate on the opc-ua-servers side.
But in order to recreate the exact same certificate, you have to provide the `clientCertificateSeed`.

#### 2.) 
```yaml
---
input:
  opcua:
    endpoint: opc.tcp://ip-address:port
    username: ''
    password: ''
    subscribeEnabled: true
    useHeartbeat: true
    securityMode: SignAndEncrypt
    securityPolicy: Basic256Sha256
    clientCertificateSeed: "1T3kZzU39FkcUHY9992koPaJ1X03EApIETMiBcjgqosDpvLjmU97ipATYRg8uxq3"
    nodeIDs:
      - i=2255
```

[basic256sha256.log](https://github.com/user-attachments/files/18733868/basic256sha256.log)


Some of the certificates items are set in a certain behaviour e.g. the `NotBefore`-Timestamp is set to the beginning of the year in order to recreate the exact same certificate each time we rerun the application.

#### Testing:
- works for Kepware
- works for Wago (automatically accepting certificates)
- Siemens still under investigation

currently preparing Ginkgo-Tests for Kepware

led0nk:
So this is basically the new user-flow when using "encrypted" connection, which will then generate a certificate and provide you with a base64-encoded string, you could then use in the input to re-use the cert.

image

opcua_plugin/opcua.go
```diff
@@ -137,6 +139,11 @@ func newOPCUAInput(conf *service.ParsedConfig, mgr *service.Resources) (service.
 		return nil, err
 	}
 
+	clientCertificate, err := conf.FieldString("clientCertificate")
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
led0nk: in accordance to the other fields this doesn't make sense here

opcua_plugin/opcua_kepware_test.go
```diff
@@ -103,6 +103,47 @@ var _ = Describe("Test against KepServer EX6", func() {
...
...FSd3pUb3grL2dNSDVteHYwT2MrVTZ0d295YzlECndEU1F2VjZFZlFWZXFpK0YwZEFVMGFGTHMxNjlWL3h3YTNUYW5IVDBIR21JWHM4NVZ4aFFPR3d4QVRMZGRZcG8KOVVDQ21QcTFIa3Z0blB6clA4NVlYQ21NcEFHYU40SDFKK2VZb0hmQjJoTUxvRUJUT3Bwc2gzRUNnWUVBeWxyMwpMTmVzU0NqaFFjVlJSM1FKSTY1UGFMaFl2bmhTL29OeFA4ZnFaMk4wd0Fwb0QxY3R5L2JWNUlId1ErYmhUWWlpCnlGcTAxMkV4dmwxSTd3K0FHSHN0V3JiM056aVB0V2ZkYXRyRnZUOUlJRldmYlhBb2ZQd0x6L0lyZnE1czVVNFEKSTBjN3A3b29xN0U1c0c0dml0ZmFBTEpQdVZsS3lqU1FEdG9TUFljQ2dZQW0zS3ErVzJQU3hsU3pkcFBERHZPcQpPZ0JPUTBUeWppczRxMGJ1NndFamloT3lkNEdnRTZuNndnNW0wYnhOMk50Z1kwc2xvTlpnbkFLQTQvZDNIQ1RkCkxpSjVtWHd5U09pK0RJUlJJaitVaWUwRE50RkRUdkJ3TXNPQ04rUEZRYXNaL08vdWNvRmlwZFNTcFRmM240REIKdEJmU3B0K3htN29ka1pSeVJiVXQ0UUtCZ1FDQ0U1ZHdLL09ETDBRZGswbDhOUXJxOU5IMjd3bWM1cHZ6SDJ0ZApKSlY1Z0dVOFRYUTI2RU40S1dPMVVCR3lsR2VmLzRVa1phcDZDUENBL2xZejFqTHhpYnpONDI5bS8rVGhKN01kClRTUnZVbzU2dW8rUk9kZk0vL2hYNDhReDJzNEZXUGptdEpPVWJnWlcxOFhOdEFhN3FhVnBiVFh5WDRQRUY2WEMKNWt0ellRS0JnUUNjeTU3aHdyLzFXQlFvSzc0akRDUmpWaXRXbVBOcytiQU9RMWtZRGp6WW5mSHhQZTVEZkxteAp0cnd1bHczOEZ1VUErelVhbmp6SCtDcXNQSHlRNUpOY1JDcStnRFlyTUxUV1B2VmVKSDUzS3JRTVZFUitzRDZ6CjltVy90VCtETUhrZHo0NGh0clBlNWdoTlRPeTBKVmV3T3c1bDBva1N6QW9SYXRhSi9KSnZqQT09Ci0tLS0tRU5EIFJTQSBQUklWQVRFIEtFWS0tLS0tCg==",
```
JeremyTheocharis, line 119: leave a comment here stating it is a local test certficiate

README.md
```diff
@@ -167,14 +167,23 @@ input:
     password: 'your-password'
 ```
 
-##### Security Mode and Security Policy
+##### Security Options
```
JeremyTheocharis, line 170: adjust README

JeremyTheocharis (approved):

PR #135, opened 2025-03-05
[ENG-2546] OPC UA Plugin Performance Optimization

kanapuli:
## Summary
This PR adjusts several key parameters in the OPC UA plugin's worker management system to optimize performance and throughput.

## Changes Made
- Increased maximum concurrent workers from 200 to 500
- Increased initial worker count from 10 to 25
- Extended target latency from 250ms to 500ms to allow more time for operations
- Adjusted worker scaling algorithm:
  - Now adds/removes 20 workers at a time (previously 10) for faster adaptation

## Motivation
These changes aim to improve the plugin's performance when dealing with OPC UA servers, particularly in environments with high latency or many concurrent operations. The higher initial worker count allows for faster startup performance, while the more aggressive scaling enables better adaptation to changing server conditions.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the system’s resource management strategy to better adapt to varying load conditions, resulting in improved responsiveness and performance.

### DIFF
--- a/opcua_plugin/browse_workers.go
+++ b/opcua_plugin/browse_workers.go
@@ -8,11 +8,11 @@ import (
 )
 
 const (
-	MaxWorkers     = 200
+	MaxWorkers     = 500
 	MinWorkers     = 5
-	InitialWorkers = 10
+	InitialWorkers = 25
 	SampleSize     = 5 // Number of requsts to measure response time
-	TargetLatency  = 250 * time.Millisecond
+	TargetLatency  = 500 * time.Millisecond
 )
 
 // ServerMetrics is a struct that holds the metrics for the OPCUA server requests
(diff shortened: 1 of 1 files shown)


opcua_plugin/browse_workers.go
```diff
@@ -8,11 +8,11 @@ import (
 )
 
 const (
-	MaxWorkers     = 200
+	MaxWorkers     = 500
 	MinWorkers     = 5
```
JeremyTheocharis, line 12: i think for sensible OPC UA servers minworkers needs to be 1, also then by adding or removeing always 20 we need to ensure that we can also land up with 1. maybe add some type of scaling in there? so the closer it is to a low latency, the more workers we can add, if its increasing too much, then also reduce the workers a lot.

the most complicated way to do it is a PID controller, but maybe you find an easier option

JeremyTheocharis (changes requested):

PR #144, opened 2025-04-02
ENG-2739: no nodes detected after reconnect/connection failure

led0nk:
- this should resolve the bug described in the linear issue
-> whenever you reconnect it previously didn't detect any nodes therefore it was simply stuck at this point
-> now it should reuse the previously stored nodes and `nodeDef` and check if it was `FullyDiscovered` and in a certain time period

- tested with opcsimv2 (since it was the only reliable source, which fails continuously) and worked 

## Summary by CodeRabbit

- **New Features**
	- Improved node discovery efficiency through enhanced tracking and caching mechanisms.
	- Streamlined processing now skips redundant operations for nodes that remain up-to-date, delivering a more responsive experience for users.
	- Introduced a mechanism to manage node freshness and discovery status, enhancing overall control flow.
	- Added a new constant for stale time management in node discovery.

### DIFF
--- a/opcua_plugin/core_browse.go
+++ b/opcua_plugin/core_browse.go
@@ -15,7 +15,9 @@ import (
 
 const (
 	MaxTagsToBrowse = 100_000
-	StaleAfter      = 15 * time.Minute
+	// StaleTime is used to define whether a node that was discovered is marked
+	// as stale to rediscover -> if maybe an attribute changed from the previous run
+	StaleTime = 15 * time.Minute
 )
 
 type NodeDef struct {
@@ -202,7 +204,7 @@ func worker(
 					vni = VisitedNodeInfo{}
 				}
 				// to skip nodes that are already FullyDiscovered and fresh
-				if vni.FullyDiscovered && time.Since(vni.LastSeen) < StaleAfter {
+				if vni.FullyDiscovered && time.Since(vni.LastSeen) < StaleTime {
 					logger.Debugf("Worker %s: node %s is fully discovered and fresh, skipping..", id, task.node.ID().String)
 					taskWg.Done()
 					continue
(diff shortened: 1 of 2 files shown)


opcua_plugin/core_browse.go
```diff
@@ -187,10 +195,32 @@ func worker(
...
+				if vni.FullyDiscovered && time.Since(vni.LastSeen) < StaleAfter {
+					logger.Debugf("Worker %s: node %s is fully discovered and fresh, skipping..", id, task.node.ID().String)
+					taskWg.Done()
+					continue
+				}
+
+				// set the node.ID() but not yet fully discovered if sth breaks, we can
+				// start over from here
+				visited.Store(task.node.ID(), VisitedNodeInfo{
+					Def:             vni.Def,
```
kanapuli, line 214: What is the use of this `Def` field in `VisitedNodeInfo` ? Seems like it is not used. 
If it is unused, lets remove it and we can simplify the code as follows ,
```go
			if val, found := visited.Load(task.node.ID()); found {
				var vni VisitedNodeInfo
				vni, _ = val.(VisitedNodeInfo)

				if vni.FullyDiscovered && time.Since(vni.LastSeen) < StaleAfter {
					logger.Debugf("Worker %s: node %s is fully discovered and fresh, skipping..", id, task.node.ID().String)
					taskWg.Done()
					continue
				}
			}

			visited.Store(task.node.ID(), VisitedNodeInfo{
				LastSeen:        time.Now(),
				FullyDiscovered: false,
			})
```

led0nk: oh it is acutally used in `buildNodeListFromCache() so we can use that do quickly build up the nodeList without discovering nodes again

kanapuli (approved):

PR #186, opened 2025-07-01
Eng 3137 add burst protection for topic browser

JeremyTheocharis:


topic_browser_plugin/raw_kafka.go
```diff
@@ -32,26 +68,38 @@ import (
...
 
+	// Create final headers map with optimal sizing
+	// Pre-size based on actual header count to avoid growth allocations
+	finalHeaders := make(map[string]string, len(tempHeaders))
+
+	// Copy from temporary map to final map
+	// This is necessary because the EventKafka object has a longer lifecycle
+	// than our pooled temporary map
+	for k, v := range tempHeaders {
+		finalHeaders[k] = v
```
Scarjit, line 92: maps.Copy

Scarjit:
bugbot run

Scarjit:
^
In our tests it is shown that the old implementation over estimated by a factor of ~2000, while our new implementation overestimated by only ~200, already giving a better estimation.

JeremyTheocharis:
From O3:
- You divide by runtime.NumCPU(). If operators pin the process with GOMAXPROCS=1, your percentage will be 25 % on a 4‑core box even when the goroutine is pegged. Divide by runtime.GOMAXPROCS(0) instead

Scarjit (approved):

PR #188, opened 2025-07-02
ENG-3099 Schema registry usage inside uns_output

Scarjit:
# Add JSON Schema Validation to UNS Output Plugin

## What Changed
- **Added comprehensive JSON schema validation** for UNS (Unified Namespace) topics
- **Integrated with Schema Registry** for centralized contract management
- **Enhanced message enrichment** with contract metadata (name, version)
- **Improved testing infrastructure** with testcontainers for reliable integration tests

## Why
Enable data contract validation in UNS pipelines to ensure message payload compliance and improve data quality across the unified namespace.

## Key Features
- 🔍 **JSON Schema Validation**: Validates message payloads against registered schemas
- 📊 **Contract Metadata**: Enriches messages with contract name and version headers
- 🛡️ **Fail-Open Design**: Continues operation when schema registry is unavailable
- ⚡ **High Performance**: 1,000+ validations/second (benchmarked)
- 🧪 **Comprehensive Testing**: Unit, integration, and performance tests

## Technical Details
- **New dependency**: `kaptinlin/jsonschema` for JSON schema validation
- **Schema registry integration**: Configurable URL with intelligent defaults
- **Message enrichment**: Adds `data_contract_name` and `data_contract_version` headers
- **Bypass handling**: Graceful degradation with bypass reason tracking

## Testing
- **Unit tests**: Mock schema registry scenarios
- **Integration tests**: Real Redpanda container with testcontainers
- **Performance benchmarks**: Validates performance targets
- **Separate test targets**: `make test-uns` (fast) vs `make test-uns-redpanda` (integration)

## Files Added
- `uns_plugin/schema_validation/` - Complete validation system (7 new files)
- Enhanced `uns_plugin/uns_output.go` with validation logic
- Updated test infrastructure and Makefile targets

## Performance
- **Single-threaded**: 273,523 ops/sec
- **Concurrent**: 622,438 ops/sec  
- **Memory efficient**: ~5KB per operation


Scarjit:
"Bug: Unsafe JSON Construction Causes Validation Failures"
I would classify this as acceptable risk, as the input is from another benthos processor, therefore likely good data and it is not meant as an absolute guard against bad data, but an best effort, given the overhead of json marshal the othe approach would need.

Scarjit:
bugbot run

Scarjit:
bugbot run

Scarjit:
I noticed an issue, due to the ability to "mix" numbers, strings, ... in a single datamodel this approach will not work safely.
Therefore we need to modify the registry to save schemas like:
 - `_pump_v1_timeseries-number`
 - `_pump_v1_timeseries-string`

and then modify the lookups and validations to consider the correct one based on incoming vpath.

Scarjit:
> I noticed an issue, due to the ability to "mix" numbers, strings, ... in a single datamodel this approach will not work safely. Therefore we need to modify the registry to save schemas like:
> 
> * `_pump_v1_timeseries-number`
> * `_pump_v1_timeseries-string`
> 
> and then modify the lookups and validations to consider the correct one based on incoming vpath.

Resolved.
Also resolved an issue where the wrong version could be selected.
We no longer use the schema registries internal versioning, but save a new schema per version.
This avoids issues with us not knowing if the v2 in the registry is actually the v2 from the config or just v1 that was inserted a 2nd time.

led0nk (changes requested):
rest looks good to me, schema looks okay

led0nk (approved):

PR #210, opened 2025-09-29
[ENG-3402]: timestamp-implementation for metadata-usage

led0nk:
### Description:

- fixes current implementation where the timestamp from `meta("timestamp_ms")` is used for the `tag_processors` timestamp_ms, which solves the same timestamp issue for batched messages
=> mostly used time-format here is time.RFC3339Nano, which worked for different opc-servers in our case e.g. Kepware/Wago/Siemens

- fixes a bad logline in opcua-logging for the `serverCertificateFingerprint`


downsampler_plugin/processor.go
```diff
@@ -274,7 +274,7 @@ func (mp *MessageProcessor) processWithState(msg *service.Message, dataMap map[s
 
 	// Check for same timestamp as last processed message
 	if !state.lastProcessedTime.IsZero() && timestamp.Equal(state.lastProcessedTime) {
-		mp.processor.logger.Warnf("Dropping message with duplicate timestamp %v for series '%s' - subsequent messages with identical timestamps are not allowed",
+		mp.processor.logger.Warnf("Dropping message with duplicate timestamp %v for series '%s' - subsequent messages with identical timestamps are not allowed. To prevent this issue for batched messages you can set the metafield 'timestamp_ms'.",
```
JeremyTheocharis, line 277: you can set `msg.meta.timestamp_ms` to, for example. msg.meta.opcua_whatever

JeremyTheocharis (changes requested):
Can you also update this in the tag_processor docs? So that we have a timestamp_ms function and that it accepts RFC3339 and unix tiemstamps in milliseconds?

JeremyTheocharis (changes requested):
as discussed: 

`msg.meta.timestamp_ms = Date.now() - (3*24*60*60*1000)`

does not seem to work

led0nk:
```
const threeHoursAgo = Date.now() - (3*60*60*1000);
msg.meta.timestamp_ms = threeHoursAgo.toString();
return msg;
```
yes, because it has to be a string for the metadata to work properly

JeremyTheocharis (approved):

PR #226, opened 2025-10-31
fix: make console.{log|info|warn|error} work more standard compliant

daniel-umh:
Both in Browsers and in NodeJS the logging functions available through console accapt variadic arguments of any type.

In the current implementation the JS runtime implicitly converts any object given to a string which results for example in "object Object" being printed for objects rather thana serialized view of the object data.

Secondly the function now accepts any number of arguments and joins them with a space separater just like nodejs does.

To be even more standards compliant the "info" method was also added which functionally is identical to the log method. This is because the go logger doesn't have a dedicated level "log".

### DIFF
--- a/docs/input/opc-ua-input.md
+++ b/docs/input/opc-ua-input.md
@@ -337,3 +337,23 @@ input:
   opcua:
     samplingInterval: 1000.0  # Sample every 1 second instead of as fast as possible
 ```
+
+## Metrics
+
+### opcua_subscription_failures_total
+
+Counter tracking OPC UA subscription failures by reason and node ID.
+
+**Labels:**
+- `reason`: Failure classification (`filter_not_allowed`, `filter_unsupported`, `node_id_unknown`, `node_id_invalid`, `other`)
+- `node_id`: OPC UA NodeID that failed to subscribe
+
+**Common Failure Reasons:**
+
+| Reason | Description | Resolution |
+|--------|-------------|------------|
+| `filter_not_allowed` | Server rejected deadband filter | Only numeric types (Int, UInt, Float, Double) support deadband |
+| `filter_unsupported` | Server doesn't support filters | Disable deadband: `deadbandType: none` |
+| `node_id_unknown` | NodeID doesn't exist | Verify NodeID configuration |
+| `node_id_invalid` | NodeID syntax error | Fix NodeID string format |
+| `other` | Other subscription errors | Check server logs |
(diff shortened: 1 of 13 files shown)


daniel-umh:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (2)
> > nodered_js_plugin/nodered_js_plugin.go (2)> `134-148`: **Use idiomatic Go naming and consider handling error types explicitly.**
> > > The stringify helper works correctly but has two areas for improvement:
> > > 
> > > 1. **Naming convention**: Go convention is camelCase for local variables/functions. Consider renaming to `stringifyValue` or `stringifyData`.
> > > 2. **Error type handling**: Error types are marshaled to JSON before checking for special string representations. Consider checking for `error` interface before JSON marshaling to ensure user-friendly error messages in logs.
> > > 
> > > Apply this diff to improve the implementation:
> > > ```diff
> > > -	stringify := func(data any) string {
> > > +	stringifyValue := func(data any) string {
> > >  		if str, ok := data.(string); ok {
> > >  			return str
> > >  		}
> > >  
> > > +		if err, ok := data.(error); ok {
> > > +			return err.Error()
> > > +		}
> > > +
> > >  		if stringer, ok := data.(fmt.Stringer); ok {
> > >  			return stringer.String()
> > >  		}
> > >  
> > >  		if d, err := json.Marshal(data); err == nil {
> > >  			return string(d)
> > >  		}
> > >  
> > >  		return fmt.Sprintf("%#v", data)
> > >  	}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Don't forget to update the reference in `format_log_msg` accordingly.
> > > 
> > > `150-156`: **Use idiomatic Go naming convention.**
> > > The function logic is correct and efficiently implemented with pre-allocated slices. However, the naming should follow Go conventions.
> > > Apply this diff:
> > > ```diff
> > > -	format_log_msg := func(data ...any) string {
> > > +	formatLogMessage := func(data ...any) string {
> > >  		buf := make([]string, 0, len(data))
> > >  		for _, d := range data {
> > > -			buf = append(buf, stringify(d))
> > > +			buf = append(buf, stringifyValue(d))
> > >  		}
> > >  		return strings.Join(buf, " ")
> > >  	}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This also assumes you rename `stringify` to `stringifyValue` as suggested earlier.
> 
> 📜 Review details

Thanks CodeRabbit. I've added the suggested handling for error types to the function.

JeremyTheocharis (approved):

daniel-umh:
i have noticed an unnecessary resource overhead by going back and forth with spreading and collecting variable arguments and removed this extra step: https://github.com/united-manufacturing-hub/benthos-umh/compare/c1d9fd8d14a3addb6903d2312021882f04a9f200..97f3605f5eeb6fe3249eb20a1f7aef8c11bd0c99

PR #230, opened 2025-11-07
[ENG-3852] Profile validation improvements + flaky test fix

JeremyTheocharis:
## Summary

Follow-up PR to #XXX addressing CodeRabbit review feedback and fixing a flaky test discovered after merge to master.

### 1. Strengthen Profile Validation (23940de)

Added compile-time validation checks to catch programming errors in profile definitions:

- ✅ Added `ProfileUnknown` constant for consistency (replacing hardcoded "unknown" strings)
- ✅ Added `MinWorkers >= 1` check (prevents nonsensical zero/negative worker configs)
- ✅ Added `MaxWorkers > 0` check (prevents zero/negative maximum worker configs)
- ✅ Documented validation architecture (compile-time panic vs runtime defense-in-depth)
- ✅ Added 3 new test cases to verify panic behavior on invalid profiles

**Rationale**: Profile values are compile-time constants (hardcoded in `server_profiles.go`), while profile selection is runtime (user config or auto-detection). Therefore, `validateProfile()` enforces compile-time constraints via `init()` panics to catch programming errors early. Runtime defense-in-depth clamping remains in `NewServerMetrics()` for user input validation.

### 2. Fix Flaky Browse Deadlock Test (9bff378)

Resolved intermittent test failure in GitHub Actions CI:

**Test**: "Browse Channel Blocking Behavior (ENG-3835) when taskChan is full and context is canceled should exit gracefully without deadlock"

**Root Cause**: Test timing race condition where 50ms sleep was insufficient for slow CI runners to initialize browse goroutines before context cancellation.

**Fix**: Increased sleep from 50ms → 200ms (4x safety margin)

**Verification**: This is NOT a deadlock bug - the implementation correctly handles context cancellation via select statements with `ctx.Done()` cases on all channel sends (lines 310-317, 390-404 in `core_browse.go`).

## Test Plan

- [x] All profile validation tests pass (including 3 new panic tests)
- [x] Browse deadlock test runs reliably with 200ms timeout
- [x] Existing profile detection and usage tests unchanged

## Related Issues

- ENG-3852: Server Auto-Optimization Profiles
- ENG-3835: Prevent node loss and deadlock in browse operation

### DIFF
--- a/opcua_plugin/server_profiles.go
+++ b/opcua_plugin/server_profiles.go
@@ -31,6 +31,56 @@ const (
 	ProfileUnknown         = "unknown"
 )
 
+// FilterCapability indicates known DataChangeFilter support status for OPC UA servers.
+//
+// This enum replaces hardcoded vendor checks with declarative profile configuration.
+// Zero value (FilterUnknown) enables trial-based discovery as sane default.
+//
+// Design philosophy (per UX Standards):
+// - Opinionated Simplicity: FilterUnknown=0 default enables automatic trial mechanism
+// - Immediate Trust: System works intelligently without requiring user configuration
+// - Progressive Power: Explicit FilterSupported/FilterUnsupported values optimize known servers
+type FilterCapability int
+
+const (
+	// FilterUnknown indicates DataChangeFilter support is unknown - trial on first batch.
+	// This is the sane default (zero value) that enables automatic capability discovery.
+	// When set, MonitorBatched will attempt filter on first subscription batch:
+	//   - If StatusBadFilterNotAllowed → cache failure, recursive retry without filter
+	//   - If success → cache success, use filter on subsequent batches
+	FilterUnknown FilterCapability = 0
+
+	// FilterSupported indicates DataChangeFilter is known to be supported - use immediately.
+	// Profile declares server supports Standard DataChange Subscription Server Facet (OPC UA Part 7).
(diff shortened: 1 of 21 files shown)


JeremyTheocharis:
@coderabbitai review

opcua_plugin/server_profiles.go
```diff
@@ -225,9 +226,33 @@ func GetProfileByName(name string) ServerProfile {
...
+// 1. COMPILE-TIME validation: init() panics on invalid profile constants (fail-fast)
+// 2. RUNTIME defense-in-depth: NewServerMetrics() clamps user input at usage points
+//
+// This validateProfile() function enforces compile-time constraints on hardcoded profiles.
+// Any panic here indicates a programming error in profile definitions, not user error.
+
 // validateProfile panics if hardcoded profile violates constraints
 // This is defensive programming - catches programming mistakes in profile definitions
 func validateProfile(p ServerProfile) {
+	if p.MinWorkers < 1 {
```
daniel-umh: Why is this camparison using `< 1` while both other cases below use `<= 0`? It is the same condition.

opcua_plugin/server_profiles.go
daniel-umh: Why are the comparisons using both < 1 and <= 0 for the same check?
This creates unnecessary confusion as on first glance it looks like different checks.
I'd prefer all three to use < 1 but either way is fine as long as it is consistent.

PR #241, opened 2025-11-24
Fix flaky GlobalWorkerPool counter test (ENG-3944)

JeremyTheocharis:
## Summary

This PR contains two changes:
1. **Fix race condition** in GlobalWorkerPool counter test causing intermittent CI failures
2. **Add dedicated unit test workflow** with opt-in pattern for OPC UA unit tests

---

### Problem
Test was spawning workers that could immediately process submitted tasks, creating a race where the `pendingTasks` counter was decremented before the test could assert on it.

**Evidence:**
- Test location: `opcua_plugin/core_browse_global_pool_test.go:1518-1533`
- Expected counter=1, actual=0

### Solution
Remove `pool.SpawnWorkers(2)` call since test is verifying submission behavior (counter increment), not processing behavior.

---

### Before (master)
- Unit tests ran unconditionally alongside integration tests
- No way to run unit tests separately
- `make test-opc` ran everything (unit + simulator + hardware tests)

### After (this PR)
- Unit tests are opt-in via `TEST_OPCUA_UNIT=true` environment variable
- Dedicated `opcua-unit.yml` workflow for fast unit tests (10min timeout)
- Clear separation between test categories

**Test Files (17 files updated):**
- Added skip logic: tests only run when `TEST_OPCUA_UNIT=true` is set
- Allows integration workflows to skip unit tests (they run in dedicated workflow)

**New Makefile targets:**
```makefile
test-opc-unit       # Unit tests only (fast, no deps)
test-opc-simulator  # Simulator tests (Docker OPC-PLC)
test-opc-kepware    # Kepware hardware tests
```

**CI Workflows:**
- NEW: `opcua-unit.yml` - Fast unit tests on all PRs
- UPDATED: `opcua-plc.yml` - Uses new Makefile targets for simulator/Kepware

---

## Testing
- Unit tests pass with `make test-opc-unit`
- Tests correctly skip without `TEST_OPCUA_UNIT` env var
- Full test suite passes

## Related
- Closes: [ENG-3944](https://linear.app/united-manufacturing-hub/issue/ENG-3944)

Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/opcua-unit.yml
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: OPC UA Unit Tests
+
+on:
+  push:
+    branches: [master, staging]
+    paths:
```
led0nk: I think it's a great idea, to have it only run for this specific plugin.
But this is IMO a workflow that should be also set to `required` and then we run into issues, because it's then required on PRs that e.g. are changing things on other plugins.

If we want to have it `required` (which i would definitely say) AND not have it fully executed everytime. 
What we could do here is e.g. use:
[paths-filter](https://github.com/dorny/paths-filter)
-> runs as first step here and checks the paths
-> every other step requires it's output to be true
-> otherwise it get's skipped

led0nk (changes requested):

led0nk (approved):

PR #242, opened 2025-11-25
[ENG-3884] add workflow for benthos-bump & set-flag in main.yaml

led0nk:
## Description:

This PR introduces a new workflow which should:
- generate the app token
- checks out `united-manufacturing-hub` repository
- replaces the `BENTHOS_UMH_VERSION` in Makefile+
- checks on the replacmenet -> if not then exit 1
- creates a PR in united-manufacturing-hub

---

### Open questions:
- is the previously App for ManagementConsole cross-repo available, because it needs write-access here
    -> contents: write and pull-requests: write
- at a certain point, we could somehow check the release-notes from `benthos-umh` and include them here, but first we need to make sure that this behaves like intended

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -311,7 +311,7 @@ jobs:
           prerelease: ${{ contains(github.ref, 'beta') || contains(github.ref, 'alpha') }}
 
       - name: Trigger version bump workflow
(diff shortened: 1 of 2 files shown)


JeremyTheocharis:
@coderabbitai review

JeremyTheocharis (approved):

PR #243, opened 2025-11-27
chore: build-binaries with hetzner-runners

led0nk:
### Description:

**changes:**

- removes unused actions e.g. `setup-node`
- simplified usage with using Depot-runners, also for multi-arch building of docker-images
- `build-binaries` still needs check when creating a release
- huge improvement in docker-build times, coming from 3-4 minutes -> 50 secs with cache used
- removes also unused and unneccessary steps e.g. `qemu` 

Build-binaries took like 5-7 mins in the first run, with cache in 2nd run like 2-3 minutes per binary instead of 45-55 for the arm64 builds.

`create-manifests` is set to required, since we removed this, because it isn't neccessary anymore, it should be turned off from branch protection.


JeremyTheocharis (commented):
maybe a good case instead to go for the temp runner approach? so to slowly deprecate the old runners lke selene and use more of the hetzner workflows?

FalconTube (approved):
Two minor questions regarding depot setup version, then good to merge!

JeremyTheocharis (approved):

PR #245, opened 2025-12-01
chore: concise test output format

daniel-umh:
Issue: ENG-3952

Use a much more concise output for Ginkgo tests. The goal is to have a better experience when tests fail. Only relevant output is printed.

Also: Let go install handle version selection based on the projects `go.mod` file instead of pinning different versions in the workflows. This is the recommended way from ginkgo.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,8 @@
 # limitations under the License.
 include ./Makefile.Common
 
-GINKGO_CMD=ginkgo
-GINKGO_FLAGS=-r --output-interceptor-mode=none --succinct -trace -p --randomize-all --cover --coverprofile=cover.profile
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@
 include ./Makefile.Common
 
 GINKGO_CMD=ginkgo
-GINKGO_FLAGS=-r --output-interceptor-mode=none --github-output -vv -trace -p --randomize-all --cover --coverprofile=cover.profile
+GINKGO_FLAGS=-r --output-interceptor-mode=none --succinct -trace -p --randomize-all --cover --coverprofile=cover.profile
```
JeremyTheocharis: can we adjust it for all repos as well

PR #246, opened 2025-12-04
chore: change runner to depot for opcua + path-filter

led0nk:
## Description:

**what changed:**
- changed workflow and workflow-file-name
- switched to depot-runners
- added the `paths-filter` action as a reusable action
    - only get's triggered on:
        - `go.mod`
        - `go.sum`
    - or:
        - changes to the specified path, which are not ONLY markdown-files
- removed unneccessary envs
- set `cancel-in-progress` to false, since we usually don't want to cancel this or e.g. a main flow when a push on feat-branch happens
- grouped together with unittests and added them also to the result-check
- then added a `result-check` job, which then checks if the tests were skipped, so it succeeds OR if anything failed and then fails
-> this then should be set to `required` and remove the `required` from:
    - `go-test-opcua-plc`
    - `go-test-kepware-plc`

Then we can proceed step by step with other plugins.


FalconTube (approved):
Looks great!
Just one minor comment, mostly to make us aware of that feature in the future.

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #247, opened 2025-12-08
chore: add linting to CI

led0nk:
## Description:
Regarding [ENG-3938](https://linear.app/united-manufacturing-hub/issue/ENG-3938/makefile-and-cicd-add-linting-automation-and-improve-developer) i had some additional time over the weekend 😅 

**This PR introduces:**

- `golangci-lint` including the config-file
- `make lint` which then installs the used golangci-lint binary to `./tools` 
- adding a workflow which then should be "required"
- fixes all existent linting issues in separated commits for better tracability

### DIFF
--- a/opcua_plugin/opcua_100k_scale_test.go
+++ b/opcua_plugin/opcua_100k_scale_test.go
@@ -102,6 +102,7 @@ var _ = Describe("100k Scale Browse Test", Label("100k_scale"), func() {
 			opcuaBrowserConsumerDone := make(chan struct{})
 			go func() {
 				for range opcuaBrowserChan { //nolint:revive
+					// Discard - browse details not needed for this test
 				}
 				close(opcuaBrowserConsumerDone)
 			}()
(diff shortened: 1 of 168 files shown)


opcua_plugin/opcua_100k_scale_test.go
```diff
@@ -101,8 +101,7 @@ var _ = Describe("100k Scale Browse Test", Label("100k_scale"), func() {
 			// Browse details are discarded - channel exists to prevent worker deadlock (legacy UI code removed)
 			opcuaBrowserConsumerDone := make(chan struct{})
 			go func() {
-				for range opcuaBrowserChan {
-					// Discard - browse details not needed for this test
+				for range opcuaBrowserChan { //nolint:revive
```
daniel-umh: add the comment behind this as it might look like an "oversight" to have this loop while it is actually to drain the channel

topic_browser_plugin/event_test.go
```diff
@@ -115,7 +116,8 @@ var _ = Describe("Event Processing", func() {
 						Expect(event.GetTs().GetStringValue().GetValue()).To(Equal("test"))
 					case proto.ScalarType_BOOLEAN:
 						Expect(event.GetTs().GetBooleanValue()).NotTo(BeNil())
-						Expect(event.GetTs().GetBooleanValue().GetValue()).To(Equal(true))
+						Expect(event.GetTs().GetBooleanValue().GetValue()).To(BeTrue())
+					default:
```
daniel-umh: what is this here for?

daniel-umh (approved):

JeremyTheocharis (approved):

PR #250, opened 2025-12-10
fix: add dns validation for hostnames in certs

led0nk:
## Description:

Fixes [ENG-3793](https://linear.app/united-manufacturing-hub/issue/ENG-3793/fix-urn-placement-in-opc-ua-certificate-san-ignition-rejects-certs)

This PR introduces:

- dns validation for x509 certificates within `opcua_plugin`
-> currently the hostname is hardcoded, but if there might be any potential changes in the future, it makes sense to validate the DNS, which then might land up in the certificate to ensure there are no issues for the OPC-server.


JeremyTheocharis (approved):
OK from product

daniel-umh (approved):

PR #253, opened 2025-12-12
chore: rmv busybox from dockerfile (not needed)

led0nk:
## Description:

Due to aikido-issue, which complained about versioning of busybox in the `Dockerfile`, this gets removed as it is not needed at all here. On top of that correct the `as` statements to not get warnings.


FalconTube (approved):
Good solution and this also already runs as non-root, which is great!

PR #256, opened 2025-12-15
fix: re-encode benthos messages without UseNumber

daniel-umh:
Fixes ENG-4021

---

This fix is for benthos inputs which receive structured (JSON) data.

The Bug is that fields within the parsed JSON are not recognized as Javascript Number. Instead they are understood as `object`. The cause for this is that `msg.AsStructured()` internally uses a JSON decoder with UseNumber enabled. This turns all Numbers into `json.Number` objects. When they are passed into Goja the Javascript understands them as plain Objects.

This solution checks if `msg.AsStructured()` returned a JSON object and re-encodes the data with standard settings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM management.umh.app/oci/library/golang:1.25.5 as build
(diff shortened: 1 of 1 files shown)


daniel-umh:
@coderabbitai review

PR #257, opened 2025-12-15
fix(sparkplug): track sequence numbers per-NODE instead of per-device (ENG-4031)

JeremyTheocharis:
## Summary

- Fix Sparkplug B sequence tracking to be per-NODE instead of per-device
- Per spec, sequence numbers are NODE-scoped (shared across NBIRTH, NDATA, DBIRTH, DDATA)
- Bug caused false "Sequence gap detected" errors when NDATA/DDATA interleaved

## Problem

When a Sparkplug node publishes interleaved messages:
```
DDATA seq=15 → NDATA seq=16 → NDATA seq=17 → DDATA seq=18
```

The old code tracked sequence per-deviceKey:
- NDATA used key `"group/node"` 
- DDATA used key `"group/node/device"`

This created separate map entries, so DDATA seq=18 was compared to DDATA seq=15, triggering a false gap.

## Solution

Added `ExtractNodeKey()` helper that extracts `"group/node"` from any device key, ensuring all messages from the same node share one sequence counter.

## Changes

| File | Change |
|------|--------|
| `sparkplug_b_core.go` | Add `ExtractNodeKey()` helper |
| `sparkplug_b_input.go` | Fix 4 process methods to use nodeKey |
| `sparkplug_b_input_test_helper.go` | Fix test helpers for new key structure |
| `node_sequence_test.go` | New TDD tests for node-scoped tracking |

## Test plan

- [x] All 241 Sparkplug tests pass
- [x] New node-scoped sequence tracking tests pass
- [x] Real sequence gap detection still works

Closes ENG-4031

### DIFF
--- a/sparkplug_plugin/node_sequence_test.go
+++ b/sparkplug_plugin/node_sequence_test.go
@@ -1,0 +1,159 @@
+//go:build !integration
+
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


JeremyTheocharis:
remove stupid tests

led0nk (approved):
looks good, had a comment regarding duplication of tests,but in #261 they get removed, so i removed it

also tested this a couple of times with ignition and seemingly works

PR #264, opened 2025-12-18
feat(schema-export): add additionalProperties false to nested schema definitions

JeremyTheocharis:
## Summary

- Add `additionalProperties: false` to nested object definitions in JSON Schema output
- Monaco YAML editor will now properly validate unknown fields inside protocol configurations
- Fixes issue where unknown fields inside protocol definitions passed silently

## Changes

1. **Plugin root objects** - now have `additionalProperties: false`
2. **Nested object fields** (e.g., `opcua.handshake`) - now have `additionalProperties: false`
3. **Array items that are objects** (e.g., `modbus.addresses[]`) - now have `additionalProperties: false`

## Test Plan

- [x] Added 3 new tests for additionalProperties validation
- [x] All 47 schema-export tests pass
- [x] Verified generated schema has additionalProperties: false at all required levels
- [ ] Regenerate schemas and update ManagementConsole (follow-up task)
- [ ] Verify in Monaco editor that unknown nested fields are flagged

## Linear Issue

https://linear.app/united-manufacturing-hub/issue/ENG-4027


led0nk (commented):
Although this looks good to  me and the generation works fine, I don't get why we need that, as there is no single case, where we have `additionalProperties: true`. 
Then I'm thinking of if this even makes sense, if there isn't a specific usecase.

I think I'd need more context on this.

Also something I overlooked here - we should think about (for the future) when creating the PR, we should simply remove the old json-schema and replace it by the new one. Otherwise we will have this bloated at a certain time.
Maybe as an idea, when creating the PR we create a `*.bak` of the previous json-schemas (in case something goes wrong) and then replace the old one.

JeremyTheocharis:
> Although this looks good to me and the generation works fine, I don't get why we need that, as there is no single case, where we have `additionalProperties: true`. Then I'm thinking of if this even makes sense, if there isn't a specific usecase.
> 
> I think I'd need more context on this.

we dont want validators in the UI allowing additional properties anywhere. so it will warn if you add a non existign field to opcua or even a non-existing protocol



> 
> Also something I overlooked here - we should think about (for the future) when creating the PR, we should simply remove the old json-schema and replace it by the new one. Otherwise we will have this bloated at a certain time. Maybe as an idea, when creating the PR we create a `*.bak` of the previous json-schemas (in case something goes wrong) and then replace the old one.

this is what @dit7ya also mentioned. however, we need to be backwards comaptbile and not simply have one schema. if soemone is using an old benthos-umh version, he should also still get the old schema. what we can think of is having it in a seperate repo though

led0nk (approved):

PR #271, opened 2026-01-22
fix: update tmc/langchaingo

daniel-umh:
Fixes ENG-4118

> Affected versions of this package are vulnerable to API key exposure through improperly handled HTTP error responses, which may inadvertently include sensitive credentials in error messages or logs.The worst case impact for these vulnerabilities can be "Sensitive information exposed in error messages".


JeremyTheocharis (approved):

led0nk (approved):
so one thing i noticed, opcua-tests should indeed run when changes to go.mod are made. please feel free and create an issue for this.

did the check locally and passed

PR #275, opened 2026-01-26
fix: opc-tests run now on go.mod/go.sum changes

led0nk:
# Description:

Fixes ENG-4168

The workflow didn't run on changes for `go.mod` `go.sum` as the filter was checking for both patterns in 1 file, which is not possible. This should be resolved via this pattern here


FalconTube (changes requested):
Test it pls, see above comment.

FalconTube (approved):
Good to go then :)

JeremyTheocharis (approved):

PR #284, opened 2026-02-13
Fix(s7comm): address naming accordingly

led0nk:
# Description:

Fixes ENG-4158

This PR adjusts the `s7comm`-addressing accordingly to not anymore be confusing. Also it fixes the correct sizing of the `DateAndTime` format, as it previously crashed due to the wrong buffer size.

| Area | Before | Now | Change |
  |------|--------|-----|--------|
  | Data Block | `DB1.DW20` | `DB1.DW20` | No change — block number required |
  | Process Input | `PE0.X0.0` | `PE.X0.0` | Block number removed |
  | Process Output | `PA0.W0` | `PA.W0` | Block number removed |
  | Merker | `MK0.X0.0` | `MK.X0.0` | Block number removed |
  | Counter | `C0.W0` | `C.W0` | Block number removed |
  | Timer | `T0.W0` | `T.W0` | Block number removed |

### DIFF
--- a/docs/input/siemens-s7.md
+++ b/docs/input/siemens-s7.md
@@ -58,18 +58,18 @@ Breaking this down:
 | `offset` | Byte position within the area | `20` in `DB1.DW20` (starts at byte 20) |
 | `extra` | Required for some types only | Bit number for `X`, string length for `S` |
 
-Only **DB** (Data Blocks) uses the block number. For all other areas there is only a single memory region in the PLC, so no block number is needed.
+Only **DB** (Data Blocks) uses a block number.
 
 ### Memory Areas
 
-| Area | Name | Description | Has blocks? | Example |
-|------|------|-------------|-------------|---------|
-| `DB` | Data Block | Main data storage — most commonly used | Yes — `DB1`, `DB2`, etc. are different blocks | `DB1.DW20` |
-| `PE` | Process Input | Physical inputs (sensors, switches) | No — only one area | `PE.B0` |
-| `PA` | Process Output | Physical outputs (actuators, relays) | No — only one area | `PA.W0` |
-| `MK` | Merker (Flags) | Internal boolean/word flags | No — only one area | `MK.W0` |
-| `C` | Counter | Hardware counters | No — only one area | `C.W0` |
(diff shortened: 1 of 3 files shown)


JeremyTheocharis (approved):
Backwards compatibility looks solid — old-format addresses (`PE0.X0.0`, `MK0.W0`, etc.) continue to work, deprecation warning is clear, DB without block number correctly rejected. DT buffer fix is correct (8 bytes, not 7). Regex change is safe (`[0-9]*` is a strict superset of `[0-9]+`, disjoint character classes with `[A-Z]+`).

Two things worth noting:

---

**F1: Deduplication edge case (minor, awareness)**

If someone has *both* `PE0.X0.0` and `PE2.X0.0` in their config, the old code treated them as different addresses (different `DBNumber`). Now both normalize to `DBNumber=0` → `reflect.DeepEqual` returns true → startup fails with `"duplicate address"`.

This is actually *more correct* — gos7 never sends `DBNumber` on the wire for non-DB areas, so `PE0.X0.0` and `PE2.X0.0` always read the same physical PLC byte. The old behavior was silently reading duplicate data. But it is a config that used to start and now won't.

Probably not worth blocking on (it's a weird config to write), but worth being aware of. Could consider a more descriptive error message like *"PE2.X0.0 conflicts with PE0.X0.0: block numbers are ignored for non-DB areas, these read the same physical address"* — up to you whether to address now or in a follow-up.

---

**F2: `s7_address` metadata change on migration (please add code comment)**

When users eventually follow the deprecation warning and migrate `PE0.W0` → `PE.W0`, the `s7_address` metadata value changes too (it stores the raw config string). Any downstream bloblang matcher like:
```
match meta("s7_address") {
  "PE0.W0" => "pressure_input"
}
```
…silently stops matching. The MC template (`s7comm.template.ts`) actually demonstrates this exact pattern.

This doesn't break on upgrade (old format still works), but it *will* break when users actively migrate their configs. **Please add a code comment near the `deprecatedAddressWarning` function** noting this — so that when the deprecation is eventually enforced (old format removed), the migration guide explicitly tells users to also update all downstream `s7_address` references in their bloblang/tag processor configs.

PR #287, opened 2026-02-19
fix(schema): fields shouldnt be required when children have defaults

led0nk:
# Description

grafik

Currently in ManagementConsole we got the issue that the editor flags multiple fields in several plugins as `required`. These fields are set to required, as we check on them if they're not set to optional and do not have a default field. 

Now we will check for the children if they already have defaults set, if so the parent-field is not anymore required as this then will get set via the defaults.

### DIFF
--- a/cmd/schema-export/exporter.go
+++ b/cmd/schema-export/exporter.go
@@ -165,6 +165,10 @@ func extractFields(configObj map[string]interface{}) map[string]FieldSpec {
 			field.Children = extractFieldsArray(childFields)
 		}
 
+		if field.Required && len(field.Children) > 0 && !hasRequiredChildren(field.Children) {
+			field.Required = false
+		}
+
 		fields[field.Name] = field
 	}
 
@@ -214,6 +218,10 @@ func extractFieldsArray(childFields []interface{}) []FieldSpec {
 		// Recursively handle nested children
 		if childFieldsNested, ok := fieldMap["children"].([]interface{}); ok && len(childFieldsNested) > 0 {
 			field.Children = extractFieldsArray(childFieldsNested)
+		}
+
+		if field.Required && len(field.Children) > 0 && !hasRequiredChildren(field.Children) {
+			field.Required = false
 		}
 
 		fields = append(fields, field)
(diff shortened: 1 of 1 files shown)


rpaland (approved):

JeremyTheocharis (approved):
Looks good — the logic is correct and the fix is well-placed. The bottom-up recursion means children are fully resolved before the parent check runs, so `hasRequiredChildren()` effectively propagates through nested structures. All 47 existing tests pass.

We ran the schema exporter before/after and confirmed **183 fields** correctly flip from `required=true` to `required=false` — all are object-type config groups (`tls`, `batching`, `backoff`, `auth`, `workarounds`, etc.) where every child has a default. Zero false positives.

**One request: please rebase onto master** — PR #286 (deprecated field marker by Rogér) merged after your branch diverged, so your branch is missing that fix. No conflict expected, just needs the rebase.

---

While reviewing the schema exporter, we also found **3 additional issues** in `json_schema.go` that are independent of this PR. Would be great to either fix them here (since you're already in the schema export code) or create separate tickets:

**1. `kind: "map"` not handled (HIGH — 74 fields)**
`benthosTypeToJSONSchemaType()` handles `kind == "array"` but has no case for `kind == "map"`. Map fields like `http_client.headers` get exported as `"type": "string"` with `"default": {}` — self-contradictory. Should be `"type": "object"` with `"additionalProperties"`. This affects UMH plugins too.

**2. Component reference types default to `"string"` (73 fields)**
Benthos types `input`, `output`, `processor`, `scanner` aren't in the `mapBenthosBaseType()` switch, so they all become `"string"`. Most visible in `batching.processors` (53 fields) where items get `"type": "string"` instead of `"type": "object"`. All 73 are in upstream Benthos plugins only (not UMH custom plugins), so lower priority.

**3. `type: "unknown"` defaults to `"string"` (34 fields)**
Dynamic fields (JWT claims, endpoint params) are `FieldTypeUnknown` in Benthos (meaning "any JSON value"), but get mapped to `"string"`. The correct JSON Schema would be an empty schema `{}` (unconstrained). Again all upstream plugins, all advanced fields — lowest priority of the three.

As a side note: we also found that deprecated fields weren't being marked, but Rogér already fixed that in PR #286 👍

PR #291, opened 2026-02-23
chore(CI): refactor workflows 

led0nk:
# Description

This PR refactors the CI accordingly to only run specific tests whenever they're needed.
Also it adds tests for:
- `downsampler_plugin`
- `streamprocessor_plugin`

It should now work like this:

1. Check on the path (specific plugin / only markdown touched )
2. check the license-header
3. lint
4. based on the path (plugin) run tests (only needed ones) - if multiple run in parallel
5. get the results -> this then gets the **only** required step for tests in CI

This should then help to reduce CI-time by a lot, also make it less "painful" to wait on tests, which are not needed at all, and don't run any tests if e.g. there are linting issues or headers missing, to simply reduce the load on our testing infrastructure.

Also it doesn't run on any pushes, only on pushes directly to master and PRs, which target master
Now the unittests should guard the plc-tests, as if there are any failures we wouldn't have any needs to run the plc-tests as well.

Fixes ENG-4459


FalconTube (changes requested):
- Create a separate ticket to pin all the action SHAs, as we also do in other repos.

Besides that, I just gave some performance improvement suggestions.
Once these are adresses, it's good to go, looks great!

FalconTube (approved):
Two more suggestions for speed, but good to go now!

JeremyTheocharis (approved):

PR #303, opened 2026-03-17
bumped RuneRoven/benthosADS to v1.0.8 with symbol download fix.

RuneRoven:


RuneRoven:
> It does look like your release has old imports for:
> 
>     * `github.com/cloudflare/circl`
> 
>     * `go.opentelemetry.io/otel/sdk`
> 
> 
> Please make sure that they at least match the current version in `benthos-umh`

The benthosADS module doesn't depend on cloudflare/circl or [go.opentelemetry.io/otel/sdk](http://go.opentelemetry.io/otel/sdk) at all — they're not in its dependency graph:
 $ cd benthosADS && go mod graph | grep -E "cloudflare/circl|[opentelemetry.io/otel/sdk@](http://opentelemetry.io/otel/sdk@)"
  (empty — no results)
And they are pinned in the benthos-umh so it should be fine.

PR #305, opened 2026-03-18
fix(opcua): handle bad-status values and ExtensionObjects gracefully without dropping the batch

### LINKED ISSUE #304: fix(opcua): Handle Custom Type gracefully
## Problem

When reading OPC UA nodes that contain custom type using the benthos-umh OPC UA input plugin, it handles them inconsistently and throws error 
```
RRO[2026-02-12T13:10:54Z] Status not OK: The extension object cannot be (de)serialized because the data type id is not recognized. StatusBadDataTypeIDUnknown (0x80110000)  @service=benthos label="" path=root.input
```

Our example custom type "CoffeeMachine" is derived from/built on top of Siemens' foundation model http://www.siemens.com/simatic-s7-opcua.

Please have a look at our example Nodeset file that contains that custom type [DB_CoffeeMachine_Example.xml](https://github.com/user-attachments/files/26054219/DB_CoffeeMachine_Example.xml)

## Root Cause

The `getBytesFromValue()` function in `opcua_plugin/core_read.go` needs better handling for ExtensionObjects that couldn't be decoded by the OPC UA client library (gopcua). This happens when:
- Custom User-Defined Types (UDTs) are not registered
- Vendor-specific binary extensions are encountered
- TypeID points to unknown types

## How to Reproduce

We've prepared a test example OPC UA NodeSet file (`DB_CoffeeMachine_Example.xml`) with custom ExtensionObjects that cannot be decoded - this is a Siemens S7 OPC UA exported from PLC with custom type (namespace: `http://CoffeeMachine`)

**Steps:**
1. Download the attached `DB_CoffeeMachine_Example.xml` file
[DB_CoffeeMachine_Example.xml](https://github.com/user-attachments/files/26054219/DB_CoffeeMachine_Example.xml)
2. Import it into Prosys OPC UA Simulator 
3. Use our attach benthos pipeline to connect to the simulator and read nodes from this namespace. 
```
input:
  opcua:
    endpoint: opc.tcp://your-prosys-simulator-ip:53530
    nodeIDs:
      - ns=4;i=12
    subscribeEnabled: true
    useHeartbeat: true
    autoReconnect: true
    reconnectIntervalInSeconds: 10
pipeline:
  processors:
    - mapping: |-
        let opcua_tag_group = metadata("opcua_tag_group")
        let opcua_attr_browsename = metadata("opcua_attr_browsename")
        root = {
          ($opcua_tag_group + "." + $opcua_attr_browsename): this
        }
        let datapointId = $opcua_tag_group + "." + $opcua_attr_browsename
        root = {
          "dataPointId": $datapointId,
          "value": this.string()
        }
    - mapping: |-
        root = this
        root.source = "Prosys Simulator"
        root.ts = timestamp_unix_milli()
output:
  stdout:
    codec: lines
```
4. Observe: The plugin may fail, skip nodes without proper logging, or produce invalid messages
## Proposed Solution

We (@batthebee and myself) have prepared a fix that:
1. ✅ Explicitly handles nil ExtensionObjects in the single-object case
2. ✅ Filters out undecodable ExtensionObjects in arrays
3. ✅ Provides clear warning logs with TypeID information
4. ✅ Gracefully skips nodes that cannot be decoded instead of failing

## Next Steps
We have a PR ready that addresses this. Would you be interested in reviewing it? We can submit it for your feedback.

dattran96:
**Problem**
- In pull mode, a single bad-status DataValue (e.g. StatusBadDataTypeIDUnknown from an undecodable Siemens UDT / custom Extension Object) was failing the entire batch read:

```
ERRO  Status not OK: ... StatusBadDataTypeIDUnknown (0x80110000)
ERRO  Read failed: status not OK: ...
ERRO  Failed to read message: status not OK: ...
```
Worse, every other valid tag in the same batch was also dropped silently. On top of that, the browse phase was throwing away legitimate data-bearing Variable nodes simply because they had HasProperty/HasComponent children, and tag-group derivation had a couple of subtle path-sanitization bugs.

***Changes***
**Unified status handling across the read pipeline**
- New statusIsBad() helper that checks BAD severity per OPC UA Part 8 (top 2 bits of StatusCode), instead of exact equality with StatusOK.
- GOOD variants (GoodClamped, GoodLocalOverride, …) and UNCERTAIN codes are no longer silently dropped.
- Applied in getBytesFromValue (pull) and MonitorBatched (subscribe) for consistency.
- Removed the duplicate StatusOK check in ReadBatchPull — getBytesFromValue is now the single source of truth for "is this value emit-worthy?". This is what actually fixes the batch-failure cascade above.
- Documented Read()'s per-result-status contract in its doc comment so future callers don't have to rediscover it.

**Browse: stop dropping Variables that happen to have children**
- Removed the "skip Variable if len(children) > 0" branch in core_browse_global_pool.go. The browse follows HierarchicalReferences, so a typed Variable with HasProperty/HasComponent metadata children is a normal data-bearing tag, not a container. Filtering of bad/undecodable values is delegated downstream to getBytesFromValue.
**Extension Objects**
- Decodable extension objects are marshalled to JSON (including arrays).
- Undecodable / nil-valued extension objects are skipped with a single warning instead of erroring.

**Tag-group derivation fixes**
- Use TrimSuffix so a tag name appearing earlier in the path isn't accidentally stripped.
- Sanitize the BrowseName before stripping, since the path is built from sanitized names.
- Handle root-level nodes correctly (path equals tag name).

**Tests**
- getBytesFromValue: focused BAD case (StatusBadNodeIDUnknown) + positive UNCERTAIN case.
- createMessageFromValue tag-group derivation: nested path, repeated tag name, root-level node, raw-vs-sanitized BrowseName.
- ExtensionObject scenarios: decodable, undecodable, mixed arrays.

**Docs**
- Rewrote the Extension Objects section in docs/input/opc-ua-input.md to reflect that browsing follows HierarchicalReferences (Organizes, HasComponent, HasProperty, HasNotifier, …), and that opaque ExtensionObject members need an explicit NodeId only when the server does not expose them.
- Removed the obsolete "Browse Hierarchical References (option until v0.5.2)" section — that flag was removed in v0.6.0.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,7 +170,7 @@ msg.meta.umh_topic = `umh.v1.${msg.meta.location_path}.${msg.meta.data_contract}
 
 ## Topic Naming Convention
 
-UMH topics follow strict validation rules defined in `pkg/umhtopic/`:
+UMH topics follow strict validation rules defined in `pkg/umh/topic/`:
 
 ### Pattern Structure
 
@@ -489,25 +489,50 @@ Field(service.NewIntField("queueSize").
 ### Core Commands
 ```bash
 make all          # Clean and build
-make target       # Build benthos binary
+make build        # Build benthos binary (outputs to tmp/bin/benthos)
 make clean        # Remove build artifacts
 make test         # Run all tests with Ginkgo
+make run CONFIG=./config/opcua-hex-test.yaml LOG_LEVEL=DEBUG  # Run locally with a config
 ```
 
 ### Plugin-Specific Tests
 ```bash
-make test-opcua              # OPC UA plugin tests
+make test-unit-opc           # OPC UA unit tests (sets TEST_OPCUA_UNIT=true)
+make test-integration-opc    # OPC UA integration tests (needs hardware env vars, see `make env`)
 make test-modbus             # Modbus plugin tests
-make test-s7                 # S7 protocol tests
+make test-s7comm             # S7 protocol tests
 make test-sparkplug          # Sparkplug B tests
 make test-eip                # Ethernet/IP tests
-make test-sensorconnect      # Sensorconnect tests
-make test-stream-processor   # Stream processor tests
-make test-tag-processor      # Tag processor tests
-make test-topic-browser      # Topic browser tests
+make test-sensorconnect      # Sensorconnect tests (runs serially)
+make test-stream-processor   # Stream processor tests (with -race)
+make test-tag-processor      # Tag processor tests (sets TEST_TAG_PROCESSOR=true)
+make test-noderedjs          # Node-RED JS tests (sets TEST_NODERED_JS=true)
+make test-topic-browser      # Topic browser tests (sets TEST_TOPIC_BROWSER=1)
 make test-downsampler        # Downsampler tests
-make test-classic-to-core    # Migration plugin tests
-make test-uns                # UNS output tests
+make test-classic-to-core    # Migration plugin tests (sets TEST_CLASSIC_TO_CORE=1)
+make test-pkg-umh-topic      # UMH topic library tests
+make test-uns                # UNS tests (excludes Redpanda-dependent tests)
+make test-uns-redpanda       # Full UNS tests (requires Docker for testcontainers)
+```
+
+### Lint & Format
+```bash
+make lint              # Run golangci-lint
+make lint-fix          # Auto-fix lint issues
+make fmt               # Format with gofumpt + gci (import ordering)
+```
+
+### Code Generation
+```bash
+make proto                              # Regenerate protobuf Go files
+make generate-schema VERSION=0.11.7     # Generate plugin schemas for ManagementConsole
+```
+
+### Fuzz & Benchmarks
+```bash
+make fuzz-stream-processor   # Run fuzz tests (press Ctrl+C to stop)
+make bench-stream-processor  # Benchmark stream processor
+make bench-pkg-umh-topic     # Benchmark topic parsing
 ```
 
 ### Utility Commands
@@ -515,8 +540,33 @@ make test-uns                # UNS output tests
 make license-check    # Verify Apache 2.0 headers
 make license-fix      # Add missing license headers
 make setup-test-deps  # Install test dependencies
+make install          # Install all dev tools (lint, fmt, license, protoc, etc.)
 make serve-pprof      # Run profiling server
+make env              # Show all test environment variables
 ```
+
+### Running a Single Test
+```bash
+# Run a single test by name pattern
+go run github.com/onsi/ginkgo/v2/ginkgo --focus "test name pattern" ./plugin_dir/...
+
+# Or use FIt/FDescribe in code to focus tests (don't commit these)
+FIt("should handle valid input", func() { ... })
+```
+
+### Test Environment Variables
+
+Many tests are gated by environment variables and silently skip without them. The `make test-*` targets set these automatically, but they matter when running tests directly with `go test` or Ginkgo.
+
+| Variable | Used by | Set by make target |
+|---|---|---|
+| `TEST_TAG_PROCESSOR=true` | tag_processor_plugin tests | `test-tag-processor` |
+| `TEST_OPCUA_UNIT=true` | opcua_plugin unit tests | `test-unit-opc` |
+| `TEST_NODERED_JS=true` | nodered_js_plugin tests | `test-noderedjs` |
+| `TEST_CLASSIC_TO_CORE=1` | classic_to_core_plugin tests | `test-classic-to-core` |
+| `TEST_TOPIC_BROWSER=1` | topic_browser_plugin tests | `test-topic-browser` |
+
+Run `make env` for the full list including hardware integration test variables.
 
 ## Testing Approach
 
@@ -527,9 +577,7 @@ make serve-pprof      # Run profiling server
   - Test real protocol connections
   - Validate end-to-end data flow
 - **YAML Validation**: Schema-based config testing
-- **Benchmarks**: Performance testing for critical paths
-  - `make bench-stream-processor`
-  - `make bench-pkg-umh-topic`
+- **Benchmarks**: Performance testing for critical paths (see Fuzz & Benchmarks section above)
 
 ### Test Patterns
 ```go
@@ -568,7 +616,7 @@ benthos-umh/
 ├── config/               # Configuration examples
 ├── docs/                 # Documentation (GitBook format)
 ├── pkg/                  # Shared packages
-│   └── umhtopic/         # UMH topic parsing
+│   └── umh/topic/        # UMH topic parsing
 ├── *_plugin/             # Protocol/processor plugins
 ├── Makefile              # Build automation
 └── Makefile.Common       # Shared Make configuration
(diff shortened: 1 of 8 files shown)


opcua_plugin/core_read.go
```diff
@@ -18,44 +18,23 @@ import (
...
-	if variant == nil || variant.Value() == nil {
+	if variant == nil {
 		g.Log.Errorf("Variant is nil")
 		return nil, ""
 	}
 
 	if !errors.Is(dataValue.Status, ua.StatusOK) {
-		g.Log.Warnf("Received bad status %v for node %s", dataValue.Status, nodeDef.NodeID.String())
+		g.Log.Warnf("Skipping node %s: status %v", nodeDef.NodeID.String(), dataValue.Status)
+		return nil, ""
```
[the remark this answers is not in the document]
led0nk: what coderabbit says here

opcua_plugin/read.go
```diff
@@ -499,6 +499,11 @@ func (g *OPCUAInput) ReadBatchPull(ctx context.Context) (service.MessageBatch, s
 			g.Log.Debugf("Received nil in item structure on node %s. This can occur when subscribing to an OPC UA folder and may be ignored.", node.NodeID.String())
 			continue
 		}
+		if value.Status != ua.StatusOK {
```
led0nk: check should only be for `StatusBad` - its the same as above to not drop values silently

opcua_plugin/core_read.go
led0nk, line 167: callers must check per-result status

opcua_plugin/core_browse_global_pool.go
```diff
@@ -744,7 +744,20 @@ func (gwp *GlobalWorkerPool) workerLoop(workerID uuid.UUID, controlChan chan str
 			// Filter by NodeClass - only send Variables to ResultChan
 			switch nodeDef.NodeClass {
 			case ua.NodeClassVariable:
-				// Variables: send to ResultChan (subscription list)
+				// Skip container/structural Variable nodes that have children.
+				// These are parent nodes (e.g., Siemens custom struct/array types like
+				// Meldungen, Betriebsmeldungen, Störmeldungen, Warnmeldungen) whose values are not meaningful on their own.
```
led0nk: try to avoid german here - also this then skips variable nodes with children, which CAN have values. We've seen this from the past and shouldn't skip that if there is a value existing

opcua_plugin/read.go
```diff
@@ -610,12 +615,14 @@ func (g *OPCUAInput) createMessageFromValue(dataValue *ua.DataValue, nodeDef Nod
...
-	// Tag Group
-	tagGroup := nodeDef.Path
-	// remove nodeDef.BrowseName from tagGroup
-	tagGroup = strings.Replace(tagGroup, nodeDef.BrowseName, "", 1)
-	// remove trailing dot
-	tagGroup = strings.TrimSuffix(tagGroup, ".")
+	// Tag Group: the path is built using sanitized browse names (see core_browse_global_pool.go),
+	// so we must remove the sanitized name from the end of the path, not the raw BrowseName.
+	// Using TrimSuffix ensures we only remove from the end, avoiding accidental mid-path matches.
+	tagGroup := strings.TrimSuffix(nodeDef.Path, "."+tagName)
```
led0nk, line 618: although this looks correct, would be nice to have this in a test

CLAUDE.md
```diff
@@ -170,7 +170,7 @@ msg.meta.umh_topic = `umh.v1.${msg.meta.location_path}.${msg.meta.data_contract}
 
 ## Topic Naming Convention
 
-UMH topics follow strict validation rules defined in `pkg/umhtopic/`:
+UMH topics follow strict validation rules defined in `pkg/umh/topic/`:
```
led0nk: would love to not have this modified at all, as there will be changes, e.g. deletion of the make commands etc. upcoming

led0nk (changes requested):
at first thank you for your contribution @dattran96 

Although we're totally fine with AI contributions, i want to point some things out.
Some findings are just out of experience, where the AI is still lacking at the current state. On top of that we often find it "bloating" the code, which is fine here. 
Therefore we always like to try to reduce code and on top of that have it tested manually (human being) as well as automatically (CI + AI). 

As you can see Coderabbit/Claude already did some comments on your PR, please go through them and check how/if they're relevant and fix those things. After working those and our comments in, we will make a CI based on this PR and manual checks. Just so that you know the process here.

opcua_plugin/read.go
dattran96: thank you :), i just added some test cases

opcua_plugin/core_read.go
dattran96: yes, i just made a change. Read() now only returns a non-nil error for session/transport failures. Per-result status check are now done by getBytesFromValue() via statusIsBad()

CLAUDE.md
dattran96: i reverted my change :)

dattran96:
@led0nk Thanks a lot for the review. I went through your comments and CodeRabbits comments, here is a summary of what changed in this push:
- **Docs**. I noticed that the documentation for OPC UA is no longer up-to-date. So i rewrote the Extension Objects paragraph and removed the obsolete "Browse Hierarchical References" section
- **Status handling unified**. Added a statusIsBad helper (BAD-severity check per OPC UA Part 8) and use it in core_read.go, read.go and read_discover.go. The duplicate StatusOK check in ReadBatchPull is removed; getBytesFromValue is now the single source of truth. This is what actually fixes the StatusBadDataTypeIDUnknown → Read failed cascade in pull mode without silently dropping GOOD/UNCERTAIN values.
- No more dropping Variables with children. Removed the len(children) > 0 skip in core_browse_global_pool.go (and the German comment) — it was dropping legitimate data-bearing Variables that happened to have HasProperty/HasComponent children. Filtering is delegated to the value-level checks downstream.
- Tag group sanitization. Kept the TrimSuffix + sanitized-name fix and added unit tests covering the three regressions: tag-name repeated in path, root-level node, and raw vs. sanitized BrowseName.
- Tests. core_read_test.go now has a focused BAD case + a positive UNCERTAIN case for getBytesFromValue, plus 4 cases for the createMessageFromValue tag-group derivation.
- CLAUDE.md reverted to master as requested.

I manually tested on my side, feel free to run your CI and your manual test whenever it suits you :smiley:

opcua_plugin/core_browse_global_pool.go
```diff
@@ -744,7 +744,20 @@ func (gwp *GlobalWorkerPool) workerLoop(workerID uuid.UUID, controlChan chan str
...
+				// Their actual data is accessed through their child nodes (e.g., [0], [1]).
+				// Only leaf Variable nodes (no children) carry real data values.
+				if len(children) > 0 {
+					if gwp.logger != nil {
+						gwp.logger.Debugf("Skipping container Variable node (has %d children) - not subscribing: nodeID=%s path=%s browseName=%s",
+							len(children), nodeDef.NodeID.String(), nodeDef.Path, nodeDef.BrowseName)
+					}
+					break
+				}
```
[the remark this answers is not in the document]
dattran96: I made a stronger fix than gating on Organizes: removed the "skip Variable if len(children) > 0" branch entirely.

Reason: in this codebase the children list comes from id.HierarchicalReferences, so even gating on Organizes wouldn't help — a perfectly valid data-bearing Variable with HasProperty/HasComponent children would still get its value dropped. Filtering is now done downstream in getBytesFromValue, which is the single place that decides whether a value is emit-worthy (BAD severity / undecodable ExtensionObject → skip; everything else → emit).

Bonus: my German comment in that block is gone too.

docs/input/opc-ua-input.md
```diff
@@ -39,12 +39,13 @@ The plugin has been rigorously tested with an array of datatypes, both as single
...
 
-There are specific datatypes which are currently not supported by the plugin and attempting to use them will result in errors. These include:
+There are specific datatypes which are currently not supported by the plugin:
 
 * Two-dimensional arrays
-* UA Extension Objects
 * Variant arrays (Arrays with multiple different datatypes)
 
+**UA Extension Objects** (custom vendor structs/UDTs) are automatically detected and skipped with a warning log. Since the browse phase discovers individual struct member nodes separately, the data within Extension Objects is still accessible by subscribing to those member nodes directly. If the Extension Object type is registered and decodable by the OPC UA library, it will be serialized as JSON.
+
```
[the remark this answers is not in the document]
dattran96: Updated the wrong old docs:
- The wording was rewritten to reflect what the current opc ua input plugin actually does in v0.11.x: it browses via id.HierarchicalReferences (which already includes Organizes, HasComponent, HasProperty, HasNotifier, etc.), so member nodes modeled with HasComponent/HasProperty are discovered.

- The note now says many ExtensionObject values are opaque (so member access requires the server to expose them via HierarchicalReferences, otherwise the user must specify the NodeId directly). I also deleted the obsolete "Browse Hierarchical References (option until v0.5.2)" section since that flag was removed in v0.6.0.

PR #308, opened 2026-03-20
[ENG-4539] Unified Address Field

rpaland:
### Summary         
  - Introduces unifiedAddresses field as a single-string alternative to the existing addresses object list   
  - Format: name.register.address.type[:key=value]* (e.g. temperature.holding.100.INT16:scale=0.1)
  - Aligns Modbus configuration style with protocols using single strings                                                
  - The legacy addresses object list continues to work with a deprecation warning                                                                                  
  - Both fields are mutually exclusive, providing both results in a clear error

### Backward compatibility                                                                                                                             
  - Existing configs using addresses work unchanged (with deprecation log warning)
  - All existing metadata fields preserved; modbus_tag_unified_address is additive


led0nk (dismissed):

modbus_plugin/modbus.go
```diff
@@ -672,9 +673,10 @@ func (m *ModbusInput) newTag(item ModbusDataItemWithAddress) (modbusTag, error)
 
 	// Initialize the field
 	f := modbusTag{
-		name:    item.Name,
-		address: item.Address,
-		length:  fieldLength,
+		name:           item.Name,
+		address:        item.Address,
+		length:         fieldLength,
+		unifiedAddress: FormatModbusAddress(item),
```
JeremyTheocharis: **nit:** `FormatModbusAddress` is called for all addresses including legacy ones. A legacy address with `Type == ""` (valid per backward-compat `case ""` in validateAddressItem) produces `name.holding.100.` — trailing dot — in `modbus_tag_unified_address`.

Fix: only set the metadata field when `item.Type != ""`.

_2 agents · DA confirmed_

JeremyTheocharis (commented):
10 review agents + a devil's advocate pass that verified each finding against the actual code. I haven't traced through all of these myself, so treat everything as "Klaus flagged this" and assess for yourself.

The 4 P0 findings (slave/slaveID docs, missing BOOL/UINT16, bit range 0-7, CI lint) are already covered by CodeRabbit, Claude bot, and Janik's comments. Me, myself and Klaus agree and have nothing to add there.

See inline comments for new findings not yet covered.

| Category | Count |
|----------|-------|
| Klaus flagged (P1) | 1 |
| Klaus flagged (P2) | 4 |
| Klaus flagged (product) | 1 |
| nits | 4 |

**P1 — Missing CHANGELOG entry:** CLAUDE.md requires a CHANGELOG.md entry under `## [Unreleased]` for every PR with user-visible changes. This PR introduces the unified address format and deprecates the legacy format, both user-visible. Add an entry under `### New Features` or `### Changed`.
_2 agents · DA confirmed_

**P2 — No integration test for unified address path:** `address_parser_test.go` has thorough unit tests for `ParseModbusAddress`, but the path through `newModbusInput` (mutual exclusion check, "both empty" error, slaveID cross-validation, dedup) has no test coverage. If someone renames the option key from `slaveID` to `slave_id`, unit tests pass but the config path breaks. Add at least one smoke test.
_2 agents · DA confirmed_

**Klaus flagged (MEDIUM) — `controller` as Advanced:** `controller` was not Advanced on master. This PR adds `.Advanced()` — was this intentional? Per CLAUDE.md, endpoint/device address fields should be Basic (required). `tcp://localhost:502` is never a correct production value. If UMH Core templates always inject `controller` (making it invisible to users), a code comment explaining that would help.
_2 agents · DA demoted to P2 (intentional per ENG-4539, but reasoning should be on record)_

---

**Product observations** (not code quality, but worth discussing before merge):

**Klaus flagged (product) — Deprecation warning lacks migration pointer:** The log warning says "Migrate to unifiedAddresses" but doesn't link to the migration docs. Adding a pointer to `docs/input/modbus.md` would help users who see the warning find the migration guide.
_1 agent · DA confirmed_

JeremyTheocharis (commented):
Re-review after Roger's 12 fix commits. 4 agents + DA verified all 31 previous comments against the current code.

**Resolution summary: 20 addressed, 4 partially, 7 not addressed.**

Roger fixed the big items: seenKeys cross-validation, bit range 0-15, BOOL/UINT16 in validOutputTypes, validateAndAppend helper extraction, swallowed parse error, empty struct returns, DescribeTable for reject tests, config spec description, error wrapping, .Examples(), FormatModbusAddress empty Type guard, dots-in-names hint, canonical form comment. Good work.

**2 items need fixing before merge:**

**P1 — `slave` vs `slaveID` in docs line 269:** The options list still says `slave=<0-255>` but the parser expects `slaveID`. The example at line 284 was correctly fixed to `slaveID=2`, but the description 15 lines above was missed. One-word fix.
_3 agents independently confirmed · DA confirmed_

**P1 — Missing CHANGELOG entry:** `CHANGELOG.md` already has an `## [Unreleased]` section (with the `uns_input` change), but nothing for the unified address format or `addresses` deprecation. The CLAUDE.md changelog guidance landed on master after your branch was cut, so you may not have seen it, but the file is active and this feature should be in it.
_2 agents · DA confirmed_

Inline comments below on a few more items.

docs/input/modbus.md
```diff
@@ -250,7 +251,52 @@ input:
...
+```
+
+**Required positional segments** (dot-separated):
+1. `name` — tag name (no dots allowed)
+2. `register` — one of: `coil`, `discrete`, `holding`, `input`
+3. `address` — numeric 0–65535
+4. `type` — one of: `BIT`, `INT8L`, `INT8H`, `UINT8L`, `UINT8H`, `INT16`, `UINT16`, `INT32`, `UINT32`, `INT64`, `UINT64`, `FLOAT16`, `FLOAT32`, `FLOAT64`, `STRING`
+
+**Optional key-value pairs** (colon-separated):
+- `slave=<0-255>` — restrict to specific slave ID (default: 0 = all)
```
JeremyTheocharis: **Klaus flagged (P1)** — Still says `slave=<0-255>` but the parser expects `slaveID=<0-255>`. The example at line 284 was correctly fixed to `slaveID=2` but this description was missed. Users following this line get `unknown option key "slave"`.

Fix: `slave=<0-255>` → `slaveID=<0-255>`

_3 agents independently confirmed · DA confirmed_

JeremyTheocharis (approved):
All re-review comments addressed. LGTM.

PR #314, opened 2026-04-08
feat(noderedjs): add cache.set/get/delete api in-memory

led0nk:
# Description

This PR introduces the first iteration of caching in `noderedjs_plugin`. 
What is done here?

- added `cache.set()`, `cache.get()`, `cache.delete()` to the nodered_js JavaScript processor
- built a `cache.Store` interface for future persistent backends (e.g. sth like SQLite)

### Example Usage:

```yaml
input:
  generate:
    count: 5
    interval: 1s
    mapping: 'root.value = random_int(min: 0, max: 100)'
pipeline:
  processors:
    - nodered_js:
        code: |
          var prev = cache.get("last") || 0;
          cache.set("last", msg.payload.value);
          msg.payload.delta = msg.payload.value - prev;
          return msg;
output:
  stdout: {}
```

Fixes ENG-4751

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Improvements
+
+- Cache API for JavaScript processors: new `cache.set(key, value)`, `cache.get(key, default)`, and `cache.delete(key)` methods for tracking state across messages. Previously, state management required complex Benthos `branch`/`request_map`/`result_map` configurations. Now you can store any JSON-compatible value (strings, numbers, objects, arrays) directly from JavaScript. `cache.get` requires a default value to prevent silent undefined bugs. Available in both `nodered_js` and `tag_processor`. Currently in-memory only (lost on restart), persistent backend planned
+
 ## [0.12.2]
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-- Cache API for JavaScript processors: new `cache.set(key, value)`, `cache.get(key, default)`, and `cache.delete(key)` methods for tracking state across messages. Previously, state management required complex Benthos `branch`/`request_map`/`result_map` configurations. Now you can store any JSON-compatible value (strings, numbers, objects, arrays) directly from JavaScript. `cache.get` requires a default value to prevent silent undefined bugs. Available in both `nodered_js` and `tag_processor`. Currently in-memory only (lost on restart), persistent backend planned
+- Cache API for JavaScript processors: new `cache.set(key, value)`, `cache.get(key)`, `cache.exists(key)`, and `cache.delete(key)` methods for tracking state across messages. Previously, state management required complex Benthos `branch`/`request_map`/`result_map` configurations. Now you can store any JSON-compatible value (strings, numbers, objects, arrays) directly from JavaScript. Use `cache.exists(key)` before `cache.get(key)` to handle missing keys. Available in both `nodered_js` and `tag_processor`. Currently in-memory only (lost on restart), persistent backend planned
 
 ## [0.12.2]
 
--- a/docs/processing/javascript-api.md
+++ b/docs/processing/javascript-api.md
@@ -1,0 +1,114 @@
+# JavaScript API Reference
+
+This page documents the global objects available in the JavaScript environment shared by the `nodered_js` and `tag_processor` processors. The engine is goja (ES5.1 with some ES6 features) — no Node.js APIs are available.
+
+## msg
+
+The message object. Contains the payload and metadata of the current message.
+
+```javascript
+msg.payload    // The message content (any JSON type)
+msg.meta       // Metadata key-value pairs (strings)
+```
+
+**Return behavior:**
+- `return msg;` — pass the message through (modified or not)
+- `return null;` or `return undefined;` — drop the message
+- `return { payload: ..., meta: ... };` — create a new message
+
+**Example:**
+```javascript
+msg.payload = msg.payload * 2;
+msg.meta.processed = "true";
+return msg;
+```
+
+## console
+
+Logging functions that write to the Benthos logger.
+
+```javascript
+console.debug(...)  // DEBUG level
+console.log(...)    // INFO level
+console.info(...)   // INFO level
+console.warn(...)   // WARN level
+console.error(...)  // ERROR level
+```
+
+Accepts multiple arguments: `console.log("value is", msg.payload.value)`
+
+## cache
+
+Key-value store for maintaining state across messages. Values persist for the lifetime of the processor. Supports any JSON-compatible type: strings, numbers, booleans, objects, arrays.
+
+```javascript
+cache.set(key, value)    // Store a value under key (string)
+cache.get(key)           // Retrieve a value, returns undefined if not found
+cache.delete(key)        // Remove a key
+```
+
+### Counter
+
+```javascript
+var count = cache.get("counter") || 0;
+count++;
+cache.set("counter", count);
+msg.payload = count;
+return msg;
+```
+
+### Previous value comparison
+
+```javascript
+var prev = cache.get("last_value");
+var delta = (typeof prev !== "undefined") ? msg.payload.value - prev : 0;
+cache.set("last_value", msg.payload.value);
+msg.payload.delta = delta;
+return msg;
+```
+
+### History (last N values)
+
+```javascript
+var history = cache.get("history") || [];
+history.push(msg.payload.value);
+if (history.length > 10) history.shift();
+cache.set("history", history);
+```
+
+### Alarm state tracking
+
+```javascript
+var alarmed = cache.get("alarm_active") || false;
+if (msg.payload > 100 && !alarmed) {
+  cache.set("alarm_active", true);
+  msg.meta.alarm = "triggered";
+  return msg;
+}
+if (msg.payload <= 100 && alarmed) {
+  cache.set("alarm_active", false);
+  msg.meta.alarm = "cleared";
+  return msg;
+}
+return msg;
+```
+
+### Configuration
+
+The cache backend and expiration can be configured in the processor YAML. By default, the cache uses the `memory` backend with no expiration.
+
+```yaml
+pipeline:
+  processors:
+    - nodered_js:
+        code: |
+          // your code
+        cache:
+          backend: memory
+          expiration: 0s    # 0s = no expiration (default)
+```
+
+### Limitations
+
+- **Memory backend only** — state is lost on process restart. A persistent backend is planned.
+- **No size limits** — the number of keys is bounded by your code. Use `cache.delete` to clean up unused keys.
--- a/docs/processing/javascript-api.md
+++ b/docs/processing/javascript-api.md
@@ -39,20 +39,30 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
 
 ## cache
 
-Key-value store for maintaining state across messages. Values persist for the lifetime of the processor. Supports any JSON-compatible type: strings, numbers, booleans, objects, arrays.
+Key-value store for maintaining state across messages. Persists across all messages for the lifetime of the Benthos process. In-memory only, lost on restart. Supports any JSON-compatible type: strings, numbers, booleans, objects, arrays.
+
+The cache is automatic and requires no configuration.
 
 ```javascript
-cache.set(key, value)    // Store a value under key (string)
-cache.get(key)           // Retrieve a value, returns undefined if not found
-cache.delete(key)        // Remove a key
+cache.set(key, value)           // Store a value under key (string)
+cache.get(key, default)         // Retrieve a value, returns default if not found
+cache.delete(key)               // Remove a key
+```
+
+`cache.get` requires two arguments. The second argument is the default value returned when the key does not exist. This prevents silent `undefined` bugs in counters, alarm flags, and other state.
+
+```javascript
+cache.get("counter", 0)         // returns 0 if missing
+cache.get("alarm_active", false)// returns false if missing
+cache.get("counter")            // TypeError — must provide default
 ```
 
 ### Counter
 
 ```javascript
-var count = cache.get("counter") || 0;
+var count = cache.get("count", 0);
 count++;
-cache.set("counter", count);
+cache.set("count", count);
 msg.payload = count;
 return msg;
 ```
@@ -60,8 +70,8 @@ return msg;
 ### Previous value comparison
 
 ```javascript
-var prev = cache.get("last_value");
-var delta = (typeof prev !== "undefined") ? msg.payload.value - prev : 0;
+var prev = cache.get("last_value", null);
+var delta = (prev !== null) ? msg.payload.value - prev : 0;
 cache.set("last_value", msg.payload.value);
 msg.payload.delta = delta;
 return msg;
@@ -70,22 +80,23 @@ return msg;
 ### History (last N values)
 
 ```javascript
-var history = cache.get("history") || [];
+var history = cache.get("history", []);
 history.push(msg.payload.value);
 if (history.length > 10) history.shift();
 cache.set("history", history);
+return msg;
 ```
 
 ### Alarm state tracking
 
 ```javascript
-var alarmed = cache.get("alarm_active") || false;
-if (msg.payload > 100 && !alarmed) {
+var alarmed = cache.get("alarm_active", false);
+if (msg.payload.value > 100 && !alarmed) {
   cache.set("alarm_active", true);
   msg.meta.alarm = "triggered";
   return msg;
 }
-if (msg.payload <= 100 && alarmed) {
+if (msg.payload.value <= 100 && alarmed) {
   cache.set("alarm_active", false);
   msg.meta.alarm = "cleared";
   return msg;
@@ -93,22 +104,19 @@ if (msg.payload <= 100 && alarmed) {
 return msg;
 ```
 
-### Configuration
+### Cycle time between events
 
-The cache backend and expiration can be configured in the processor YAML. By default, the cache uses the `memory` backend with no expiration.
-
-```yaml
-pipeline:
-  processors:
-    - nodered_js:
-        code: |
-          // your code
-        cache:
-          backend: memory
-          expiration: 0s    # 0s = no expiration (default)
+```javascript
+var lastMs = cache.get("last_event_ms", null);
+if (lastMs !== null) {
+  msg.payload.cycle_time_ms = Date.now() - lastMs;
+}
+cache.set("last_event_ms", Date.now());
+return msg;
 ```
 
 ### Limitations
 
-- **Memory backend only** — state is lost on process restart. A persistent backend is planned.
-- **No size limits** — the number of keys is bounded by your code. Use `cache.delete` to clean up unused keys.
+- **In-memory only** — state is lost when the Benthos process restarts. A persistent backend is planned.
+- **No size limits** — the cache grows unboundedly if keys are never deleted. Use `cache.delete` to clean up unused keys. A memory safeguard (threshold, eviction) is planned.
+- **Cache scope in `tag_processor`** — the cache is shared across all stages (`defaults`, `conditions`, `advancedProcessing`). A value set in `defaults` is visible in `advancedProcessing` within the same message.
--- a/docs/processing/javascript-api.md
+++ b/docs/processing/javascript-api.md
@@ -44,23 +44,27 @@ Key-value store for maintaining state across messages. Persists across all messa
 The cache is automatic and requires no configuration.
 
 ```javascript
-cache.set(key, value)           // Store a value under key (string)
-cache.get(key, default)         // Retrieve a value, returns default if not found
-cache.delete(key)               // Remove a key
+cache.set(key, value)    // Store a value under key (string)
+cache.get(key)           // Retrieve a value, logs error if key not found
+cache.exists(key)        // Returns true if key exists, false otherwise
+cache.delete(key)        // Remove a key
 ```
 
-`cache.get` requires two arguments. The second argument is the default value returned when the key does not exist. This prevents silent `undefined` bugs in counters, alarm flags, and other state.
+Always use `cache.exists(key)` before `cache.get(key)` to avoid error logs on missing keys.
 
 ```javascript
-cache.get("counter", 0)         // returns 0 if missing
-cache.get("alarm_active", false)// returns false if missing
-cache.get("counter")            // TypeError — must provide default
+if (cache.exists("counter")) {
+  var count = cache.get("counter");
+} else {
+  var count = 0;
+}
 ```
 
 ### Counter
 
 ```javascript
-var count = cache.get("count", 0);
+var count = 0;
+if (cache.exists("count")) { count = cache.get("count"); }
 count++;
 cache.set("count", count);
 msg.payload = count;
@@ -70,7 +74,7 @@ return msg;
 ### Previous value comparison
 
 ```javascript
-var prev = cache.get("last_value", null);
+var prev = cache.exists("last_value") ? cache.get("last_value") : null;
 var delta = (prev !== null) ? msg.payload.value - prev : 0;
 cache.set("last_value", msg.payload.value);
 msg.payload.delta = delta;
@@ -80,7 +84,7 @@ return msg;
 ### History (last N values)
 
 ```javascript
-var history = cache.get("history", []);
+var history = cache.exists("history") ? cache.get("history") : [];
 history.push(msg.payload.value);
 if (history.length > 10) history.shift();
 cache.set("history", history);
@@ -90,7 +94,7 @@ return msg;
 ### Alarm state tracking
 
 ```javascript
-var alarmed = cache.get("alarm_active", false);
+var alarmed = cache.exists("alarm_active") ? cache.get("alarm_active") : false;
 if (msg.payload.value > 100 && !alarmed) {
   cache.set("alarm_active", true);
   msg.meta.alarm = "triggered";
@@ -107,7 +111,7 @@ return msg;
 ### Cycle time between events
 
 ```javascript
-var lastMs = cache.get("last_event_ms", null);
+var lastMs = cache.exists("last_event_ms") ? cache.get("last_event_ms") : null;
 if (lastMs !== null) {
   msg.payload.cycle_time_ms = Date.now() - lastMs;
 }
--- a/docs/processing/javascript-api.md
+++ b/docs/processing/javascript-api.md
@@ -74,8 +74,14 @@ return msg;
 ### Previous value comparison
 
 ```javascript
-var prev = cache.exists("last_value") ? cache.get("last_value") : null;
-var delta = (prev !== null) ? msg.payload.value - prev : 0;
+var prev = null;
+if (cache.exists("last_value")) {
+  prev = cache.get("last_value");
+}
+var delta = 0;
+if (prev !== null) {
+  delta = msg.payload.value - prev;
+}
 cache.set("last_value", msg.payload.value);
 msg.payload.delta = delta;
 return msg;
@@ -84,7 +90,10 @@ return msg;
 ### History (last N values)
 
 ```javascript
-var history = cache.exists("history") ? cache.get("history") : [];
+var history = [];
+if (cache.exists("history")) {
+  history = cache.get("history");
+}
 history.push(msg.payload.value);
 if (history.length > 10) history.shift();
 cache.set("history", history);
@@ -94,7 +103,10 @@ return msg;
 ### Alarm state tracking
 
 ```javascript
-var alarmed = cache.exists("alarm_active") ? cache.get("alarm_active") : false;
+var alarmed = false;
+if (cache.exists("alarm_active")) {
+  alarmed = cache.get("alarm_active");
+}
 if (msg.payload.value > 100 && !alarmed) {
   cache.set("alarm_active", true);
   msg.meta.alarm = "triggered";
@@ -111,7 +123,10 @@ return msg;
 ### Cycle time between events
 
 ```javascript
-var lastMs = cache.exists("last_event_ms") ? cache.get("last_event_ms") : null;
+var lastMs = null;
+if (cache.exists("last_event_ms")) {
+  lastMs = cache.get("last_event_ms");
+}
 if (lastMs !== null) {
   msg.payload.cycle_time_ms = Date.now() - lastMs;
 }
--- a/nodered_js_plugin/cache/cache.go
+++ b/nodered_js_plugin/cache/cache.go
@@ -22,8 +22,6 @@ type Cache interface {
 	Set(ctx context.Context, key string, value any) error
 	// Get returns the value stored under key and if it even exists.
 	Get(ctx context.Context, key string) (any, bool)
-	// Exists returns true if key is present and not expired.
-	Exists(ctx context.Context, key string) bool
 	// Delete removes the entry for key. No-op when key does not exist.
 	Delete(ctx context.Context, key string) error
 	// Close releases any resources held by the store.
--- a/nodered_js_plugin/cache/cache.go
+++ b/nodered_js_plugin/cache/cache.go
@@ -22,6 +22,8 @@ type Cache interface {
 	Set(ctx context.Context, key string, value any) error
 	// Get returns the value stored under key and if it even exists.
 	Get(ctx context.Context, key string) (any, bool)
+	// Exists returns true if key is present and not expired.
+	Exists(ctx context.Context, key string) bool
 	// Delete removes the entry for key. No-op when key does not exist.
 	Delete(ctx context.Context, key string) error
 	// Close releases any resources held by the store.
--- a/nodered_js_plugin/cache/memory.go
+++ b/nodered_js_plugin/cache/memory.go
@@ -14,39 +14,129 @@
 
 package cache
 
-import "sync"
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+// Item wraps a cached value with an optional expiration timestamp.
+type Item struct {
+	Value      any
+	Expiration int64 // UnixNano; 0 means no expiration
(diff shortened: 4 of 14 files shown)


nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -468,7 +494,7 @@ Message content: %v`,
 
 // Close gracefully shuts down the processor.
 func (u *NodeREDJSProcessor) Close(_ context.Context) error {
-	return nil
+	return u.cache.Close()
 }
```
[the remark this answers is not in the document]
led0nk: this is tracked under eng-4768

nodered_js_plugin/cache/store.go
```diff
@@ -0,0 +1,27 @@
...
+type Store interface {
+	// Set stores value under key, overwriting any existing entry.
+	Set(key string, value any)
+	// Get returns the value stored under key and if it even exists.
+	Get(key string) (any, bool)
+	// Delete removes the entry for key. No-op when key does not exist.
+	Delete(key string)
+	// Close releases any resources held by the store.
+	Close() error
+}
```
FalconTube: While the current implementation of the store with only in-memory can not fail for `Set` and `Delete`, it can throw errors if we later switch to a permanent store on disk.
This small change here will make implementing this in the future easier.
For now, you can just return `nil` from the in-memory functions and you satisfy the interface.
```suggestion
type Store interface {
	// Set stores value under key, overwriting any existing entry.
	Set(key string, value any) error
	// Get returns the value stored under key and if it even exists.
	Get(key string) (any, bool)
	// Delete removes the entry for key. No-op when key does not exist.
	Delete(key string) error
	// Close releases any resources held by the store.
	Close() error
}
```

FalconTube (changes requested):
Looks great!
Two points:
1. Add the `error` returns to the interface
2. Cache data usually have some form of `expiration`. Our current solution does not. Not sure, if you want to think about this now, or want to do this in a follow-up.
You would probably attach an expiration duration to each KV-item.
Then a long-running goroutine with a ticker checks the values regulary and performs clean up of expired entries.
Simple implementation of this is for example here: https://github.com/patrickmn/go-cache/blob/master/cache.go#L1076

nodered_js_plugin/cache/store.go
led0nk: 1. y
2. let's do the expiration now and a potential size-limitation in a follow-up

nodered_js_plugin/cache/memory.go
```diff
@@ -0,0 +1,142 @@
...
+func NewMemoryStore(defaultExpiration time.Duration) *MemoryStore {
+	m := &MemoryStore{
+		items:             make(map[string]Item),
+		defaultExpiration: defaultExpiration,
+	}
+	if defaultExpiration > 0 {
+		j := newJanitor(1 * time.Hour)
+		m.janitor = j
+		go j.run(m)
+	}
```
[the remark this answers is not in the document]
led0nk: this is okay for this purpose as we don't plan with items less than an hour of storage-time

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -501,28 +534,81 @@ console.log("Message metadata:", msg.meta);
...
+var alarmed = cache.get("alarm_active") || false;
+if (msg.payload > 100 && !alarmed) {
+  cache.set("alarm_active", true);
+  msg.meta.alarm = "triggered";
+  return msg;
+}
+if (msg.payload <= 100 && alarmed) {
+  cache.set("alarm_active", false);
+}
+return msg;`)).
```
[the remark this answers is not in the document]
led0nk: tracked in eng-4780

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -501,28 +534,81 @@ console.log("Message metadata:", msg.meta);
...
+		return nil, fmt.Errorf("failed to parse cache expiration: %w", err)
+	}
 
+	var c cache.Cache
+	switch backend {
+	case "memory":
+		c = cache.NewMemoryStore(expiration)
+	default:
+		return nil, fmt.Errorf("unsupported cache backend: %q (supported: memory)", backend)
+	}
```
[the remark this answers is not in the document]
led0nk: is documented and should work like 0s equals no expiration, negative not allowed

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -298,25 +302,58 @@ func stringify(data any, depth uint8) (string, error) {
...
+				return goja.Undefined()
+			}
+			return v
+		},
+		"delete": func(key string) {
+			err := u.cache.Delete(key)
+			if err != nil {
+				u.logger.Errorf("cache.delete failed: %v", err)
+			}
+		},
```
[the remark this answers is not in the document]
led0nk: tracked in eng-4752 as we usually don't use multiple threads (since now)

nodered_js_plugin/cache/cache.go
```diff
@@ -0,0 +1,27 @@
...
+type Cache interface {
+	// Set stores value under key, overwriting any existing entry.
+	Set(key string, value any) error
+	// Get returns the value stored under key and if it even exists.
+	Get(key string) (any, bool)
+	// Delete removes the entry for key. No-op when key does not exist.
+	Delete(key string) error
+	// Close releases any resources held by the store.
+	Close() error
+}
```
JeremyTheocharis, line 29: **Naming**: `cache` implies temporary storage that speeds up expensive operations — data you can rebuild from somewhere else. But this is primary state: alarm flags, counters, previous values. If the data disappears, behavior is wrong (alarms re-fire, counters reset), not just slower. That's a store, not a cache. `store.set/get/delete` is more accurate. Worth aligning on this before merge — once the name is in the JS API, it's hard to change.

**Remove TTL entirely**: A store persists. No use case for the target audience requires automatic expiration:
- Alarm flags: permanent until explicitly cleared
- Counters: accumulates, permanent
- Previous values: overwritten each message
- Sliding windows: user manages array size in JS
- "Ignore old data": user checks timestamps in JS (`if (Date.now() - entry.ts > threshold) store.delete(key)`)

Removing TTL also removes the janitor goroutine, the `Item.Expiration` field, the `cache.expiration` config field, and a class of silent-data-loss bugs (48h default = alarm state disappears without warning). The MemoryStore becomes `map[string]any` + `sync.RWMutex`.

I know Yannic requested expiration, but the naming change reframes the feature: caches expire, stores persist. If users need TTL later, it's a backwards-compatible addition.

Interface:
```go
type Store interface {
    Set(key string, value any) error
    Get(key string) (any, bool)
    Delete(key string) error
    Close() error
}
```

nodered_js_plugin/cache/cache.go
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+// Cache is used as the caching interface for nodered_js.
+type Cache interface {
```
JeremyTheocharis, line 20: Consider: `context.Context` on the methods. Every storage interface in umh-core and Benthos's `service.Cache` takes context. For in-memory this is a no-op, but a file backend needs cancellation. Adding it now means Phase 2 is a backend swap, not an interface break.

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -468,17 +505,17 @@ Message content: %v`,
```
JeremyTheocharis, line 509: `ProcessBatch(_ context.Context, ...)` discards the pipeline context. This should be `ProcessBatch(ctx context.Context, ...)`. Even without cache changes, discarding a pipeline context breaks cancellation semantics for any future I/O in this processor.

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -501,28 +538,84 @@ console.log("Message metadata:", msg.meta);
...
 msg.meta.count = (msg.meta.count || 0) + 1;
-return msg;`))
+return msg;
 
+// Example 7: Persistent counter across messages using cache
+var count = cache.get("count") || 0;
+count++;
+cache.set("count", count);
+msg.payload = count;
+return msg;
```
JeremyTheocharis, line 553: Example 8: when `msg.payload <= 100 && alarmed`, the alarm flag clears but msg returns with no metadata. Users copying this will get silent de-alarms.

Consider:
```javascript
if (msg.payload <= 100 && alarmed) {
  store.set("alarm_active", false);
  msg.meta.alarm = "cleared";
  return msg;
}
return null; // No state change
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Improvements
```
JeremyTheocharis, line 5: nit: says "persistent SQLite backend planned." The current plan is a file-based backend, not necessarily SQLite. Consider "persistent backend planned" to keep options open.

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -298,25 +302,58 @@ func stringify(data any, depth uint8) (string, error) {
```
JeremyTheocharis, line 302: `docs/processing/node-red-javascript-processor.md` has no mention of the new API. For this to be shippable, users need to find the feature.

Two docs additions:

1. A dedicated `docs/processing/javascript-api.md` covering all shared JS globals (`msg`, `console`, `store`). Both nodered_js and tag_processor use SetupJSEnvironment, so the API is shared. This avoids duplicating docs across two processor pages. Similar to how Benthos has bloblang as its own docs section. Add to `SUMMARY.md` under Processing.

2. Cross-references from `node-red-javascript-processor.md` and `tag-processor.md` to the new JS API page.

Follow-up (MC repo): autocomplete in the code editor, link to docs from editor.

tag_processor_plugin/tag_processor_plugin.go
```diff
@@ -161,8 +162,9 @@ type TagProcessor struct {
 }
 
 func newTagProcessor(config TagProcessorConfig, logger *service.Logger, metrics *service.Metrics) (*TagProcessor, error) {
-	// Create a NodeREDJSProcessor for SetupJSEnvironment helper
-	jsProcessor, err := nodered_js_plugin.NewNodeREDJSProcessor("", logger, metrics)
+	// TagProcessor only uses the JS processor for SetupJSEnvironment, not for caching.
+	// In-memory with no expiration is sufficient here.
```
JeremyTheocharis, line 166: TagProcessor.Close() at line 680 still returns nil without calling jsProcessor.Close(). ENG-4768. If TTL/janitor are removed this is less urgent (Close on MemoryStore would be a no-op). But still matters for a future file backend.

nodered_js_plugin/nodered_js_plugin.go
JeremyTheocharis: actually i meant the example below :)

nodered_js_plugin/cache/cache.go
JeremyTheocharis: for now, you could leave the implementation like this (no need to throw away good code) and only change how the user uses it. We could just gradually expose it by starting wihtout TTL and only when we see people want to, deliver it in a second step

JeremyTheocharis: also i am just seeing that the code seems still be in the process of being implemented, so ttl exists in the in memory thing with janitor, but not liek that in the itnerface. i would keep it like that.

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -501,28 +538,84 @@ console.log("Message metadata:", msg.meta);
...
+}
+return msg;`)).
+	Field(service.NewObjectField("cache",
+		service.NewStringField("backend").
+			Description("Cache backend to use.").
+			Default("memory").
+			Examples("memory"),
+		service.NewDurationField("expiration").
+			Description("Default expiration duration for cached items. Items are automatically removed after expiration. Set to 0s to disable expiration.").
+			Default("48h"),
```
JeremyTheocharis: default should be 0 / endless

nodered_js_plugin/cache/cache.go
led0nk: We discussed this and decided to keep `cache` for now. The naming aligns with how users think about it — `cache.set`/`cache.get` is the familiar pattern from Node-RED context storage and web APIs (localStorage, sessionStorage). Calling it `store` would suggest durability that the memory backend does not provide. When we add a persistent backend, the name still works — Redis and Memcached are called "caches" despite being primary state in many systems. The JS API surface is stable either way — renaming later is a one-word change in the bindings if needed.

nodered_js_plugin/cache/cache.go
led0nk: Done — `Set`, `Get`, `Delete` all take `context.Context` now. MemoryStore ignores it with `_`, future backends can use it for cancellation.

nodered_js_plugin/nodered_js_plugin.go
led0nk: Fixed — `ctx` is now threaded through the entire call chain: `ProcessBatch(ctx)` → `processSingleMessage(ctx)` → `SetupJSEnvironment(ctx)` → `setupCache(ctx)`. Applied to both `NodeREDJSProcessor` and `TagProcessor`.

nodered_js_plugin/nodered_js_plugin.go
led0nk: Fixed — added `msg.meta.alarm = "cleared"` and explicit `return msg` in the de-alarm branch.

CHANGELOG.md
led0nk: Fixed — changed to "persistent backend planned".

nodered_js_plugin/nodered_js_plugin.go
led0nk: Created `docs/processing/javascript-api.md` covering `msg`, `console`, and `cache` with usage examples. Cross-referenced from both `node-red-javascript-processor.md` and `tag-processor.md`.

tag_processor_plugin/tag_processor_plugin.go
led0nk: Tracked in ENG-4768.

nodered_js_plugin/cache/cache.go
led0nk: Kept the TTL implementation internally (Item struct, janitor, closeOnce) but changed the default to `0s` — no expiration out of the box. Alarm flags, counters, and previous values persist as expected. Expiration is opt-in via config for users who need it.

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,114 @@
...
+history.push(msg.payload.value);
+if (history.length > 10) history.shift();
+cache.set("history", history);
+```
+
+### Alarm state tracking
+
+```javascript
+var alarmed = cache.get("alarm_active") || false;
+if (msg.payload > 100 && !alarmed) {
```
JeremyTheocharis: Klaus flagged. After `tag_processor` the payload is `{value, timestamp_ms}`, so `msg.payload > 100` compares an object to a number and never fires. The Previous-value and History examples on this page already use `msg.payload.value`. Same fix also needed at line 88 in this file and at `nodered_js_plugin/nodered_js_plugin.go:551,556` (Example 8 in NewConfigSpec) — I've left a pointer there.

```suggestion
if (msg.payload.value > 100 && !alarmed) {
```

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,114 @@
...
+  cache.set("alarm_active", true);
+  msg.meta.alarm = "triggered";
+  return msg;
+}
+if (msg.payload <= 100 && alarmed) {
```
JeremyTheocharis: Same fix: `msg.payload.value <= 100`.

```suggestion
if (msg.payload.value <= 100 && alarmed) {
```

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -501,28 +537,86 @@ console.log("Message metadata:", msg.meta);
...
+// Example 7: Persistent counter across messages using cache
+var count = cache.get("count") || 0;
+count++;
+cache.set("count", count);
+msg.payload = count;
+return msg;
+
+// Example 8: Alarm state that only fires once per active condition
+var alarmed = cache.get("alarm_active") || false;
+if (msg.payload > 100 && !alarmed) {
```
JeremyTheocharis: Same `msg.payload.value` fix as flagged on `javascript-api.md:83`. Both files were introduced in this PR and carry the same bug — worth fixing in one pass. Apply the same correction on line 556 here as well.

```suggestion
if (msg.payload.value > 100 && !alarmed) {
```

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,114 @@
...
+console.info(...)   // INFO level
+console.warn(...)   // WARN level
+console.error(...)  // ERROR level
+```
+
+Accepts multiple arguments: `console.log("value is", msg.payload.value)`
+
+## cache
+
+Key-value store for maintaining state across messages. Values persist for the lifetime of the processor. Supports any JSON-compatible type: strings, numbers, booleans, objects, arrays.
```
JeremyTheocharis: "Lifetime of the processor" is ambiguous — a reader thinking of processors as pipeline stages may read it as per-message. Your CHANGELOG wording is more precise and surfaces the restart caveat up-front, which is the key thing users need to know before they reach the examples.

```suggestion
Key-value store for maintaining state across messages. Persists across all messages for the lifetime of the Benthos process. In-memory only, lost on restart. Supports any JSON-compatible type: strings, numbers, booleans, objects, arrays.
```

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+### History (last N values)
+
+```javascript
+var history = cache.get("history") || [];
+history.push(msg.payload.value);
+if (history.length > 10) history.shift();
+cache.set("history", history);
+```
```
JeremyTheocharis, line 101: This snippet silently drops the message — the History example stores the value but never returns. Either add `return msg;` or a comment explaining the drop is intentional.

```suggestion
cache.set("history", history);
return msg;
```

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,114 @@
...
+  msg.meta.alarm = "triggered";
+  return msg;
+}
+if (msg.payload <= 100 && alarmed) {
+  cache.set("alarm_active", false);
+  msg.meta.alarm = "cleared";
+  return msg;
+}
+return msg;
+```
```
JeremyTheocharis, line 121: Nit: the rollout ticket's discoverability checklist names counter + alarm + cycle-time, and only two land. A brief cycle-time example here would round out the set. Suggested copy (add it as a new `### Cycle time between events` subsection before `### Configuration`; uses the two-arg `cache.get(key, default)` shape from Blocker 4 — adjust to whichever shape we pick):

    var lastMs = cache.get("last_event_ms", null);
    if (lastMs !== null) {
      msg.payload.cycle_time_ms = Date.now() - lastMs;
    }
    cache.set("last_event_ms", Date.now());
    return msg;

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,114 @@
...
+          // your code
+        cache:
+          backend: memory
+          expiration: 0s    # 0s = no expiration (default)
+```
+
+### Limitations
+
+- **Memory backend only** — state is lost on process restart. A persistent backend is planned.
+- **No size limits** — the number of keys is bounded by your code. Use `cache.delete` to clean up unused keys.
```
JeremyTheocharis: Two things in the Limitations block:

1. **Memory warning underestimates the risk.** A user building a 10k-entry history across thousands of tags won't know they're approaching OOM territory. Once the memory-safeguard follow-up ticket is filed under ENG-3463, link it here and call out the risk more plainly.
2. **Cache scope across `tag_processor` stages is not documented.** Is state shared between `defaults`, `conditions`, and `advancedProcessing`? Your code calls `cache.NewMemoryStore(0)` once per tag_processor instance, so it looks like yes — but the doc should say it plainly. Matters for anyone building cross-stage state.

```suggestion
- **Memory backend only** — state is lost on process restart. A persistent backend is planned.
- **No memory cap today** — total stored data is bounded only by your code. Avoid large arrays or high-cardinality key spaces; there is currently no OOM safeguard. A configurable memory limit is tracked under ENG-3463 as a follow-up ticket.
- **Scope** — within a single `tag_processor`, the cache is shared across all stages (`defaults`, `conditions`, `advancedProcessing`). Different processor instances get different caches.
```

JeremyTheocharis (dismissed):
Two product asks in this re-review (Blockers 1 and 4 below) plus a handful of smaller fixes in inline comments. Blocker 1 shrinks the config surface. Blocker 4 reshapes the JS `cache.get` API — I've got a pick but want your read before you code it.

## 1. Strip the user-facing cache config from `nodered_js`

Our UX standards say we want sane defaults. Right now there's no case where a user needs to adjust the cache backend or expiration, and when we eventually introduce a persistent store, we'll probably flip it on for everyone rather than making it a per-pipeline toggle. TTL should be configured in the code itself — if a user wants expiration for a specific key, the right shape is `cache.set("k", value, ttl)` with an optional third argument, not a global YAML knob applied to all keys in the processor. A separate config block beside the code also pulls a lot of UI work with it (a second form surface in the processing tab, validation, wire-format changes). Not worth exposing any of that now for zero current customer need.

Concretely: `nodered_js_plugin.go:562-573` exposes `cache.backend` and `cache.expiration` as Benthos YAML fields. `tag_processor_plugin.go:167` already does the right thing — hard-codes `NewMemoryStore(0)`, no knobs. Same PR, same cache, two philosophies. Let's match the simpler one.

- Remove the `cache` ObjectField from `NodeREDJSConfigSpec`
- Remove the `FieldString("cache", "backend")` and `FieldDuration("cache", "expiration")` parsing in Init
- Construct with `cache.NewMemoryStore(0)`, same as `tag_processor`
- Drop the `### Configuration` section from `javascript-api.md`; replace with one line saying the cache is automatic
- Audit `CHANGELOG.md` for any line that advertises configurability

The internal TTL/janitor plumbing in `MemoryStore` stays. That was my Apr 10 walk-back: keep it internal, don't expose. This ask is about the Benthos YAML surface, not the Go interface.

## 2. `SUMMARY.md` entry + verify gitbook rendering

Common failure pattern — the new `javascript-api.md` page isn't in the gitbook TOC so the sidebar won't show it. Add one line under Processing in `docs/SUMMARY.md`:

```
  - [Node-RED JavaScript Processor](processing/node-red-javascript-processor.md)
  - [JavaScript API Reference](processing/javascript-api.md)
```

After pushing, check the rendered gitbook preview. Entry looks fine in markdown but the indent/path/nesting still breaks the sidebar silently often enough that a visual confirm is worth 30 seconds.

## 4. Force the user to handle missing keys — one of two shapes

Picture the user writing a counter or an alarm rule in `nodered_js`. They're thinking about their business logic, not about JavaScript type coercion or "what happens on the first message when the cache is empty." They write `cache.get("counter") + 1`, it returns `NaN`, the processor errors, and they find out at 03:00 from a customer. We had this exact class of bug repeatedly when building the SAP DM write-flow templates in VE — same root cause every time: the cache returns undefined, the user forgets to guard, and the miss surfaces as a processor error in production. The current docs pattern (`cache.get("k") || 0`) doesn't help enough: `|| default` also silently coerces legitimate `0` / `false` / `""` / `null` back to the default, hiding genuine misses and resetting counters that roll through zero.

The API should force the user to handle the miss case at authoring time, not hope they'll remember. Two shapes can do this. I'd pick **A**, but want your read before you code it.

**Option A — mandatory default argument** (my pick):

```javascript
cache.get("counter", 0)          // 0 if missing, real value otherwise
cache.get("alarm_active", false)
// cache.get("counter")           // TypeError: cache.get(key, default) requires 2 arguments
```

```go
// ~8 lines in setupCache (nodered_js_plugin.go:334)
"get": func(call goja.FunctionCall) goja.Value {
    if len(call.Arguments) < 2 {
        panic(vm.NewTypeError("cache.get(key, default) requires 2 arguments. " +
            "Returning undefined on missing keys has caused processor failures in production. " +
            "Pass an explicit default, e.g. cache.get("counter", 0)."))
    }
    key := call.Arguments[0].String()
    v, ok := u.cache.Get(ctx, key)
    if !ok { return call.Arguments[1] }
    return vm.ToValue(v)
},
```

Usage collapses to one line per read. Familiar from Python `dict.get`, lodash `_.get`, Ruby `Hash.fetch`. Goja panics surface as real JS `TypeError` with user's line number, so forgetting the default is loud at authoring time.

**Option B — object return with `.value` / `.exists`** (alternative):

```javascript
var r = cache.get("counter");
var count = r.exists ? r.value : 0;
cache.set("counter", count + 1);
```

```go
// ~6 lines in setupCache
"get": func(key string) map[string]any {
    v, ok := u.cache.Get(ctx, key)
    if !ok { return map[string]any{"exists": false, "value": nil} }
    return map[string]any{"exists": true, "value": v}
},
```

Named fields are self-documenting, mirrors the Go interface (`Get(ctx, key) (any, bool)`) faithfully. Cost: two-line ceremony per read, and users can write `cache.get("k").value` without checking `.exists`, which silently returns `undefined` on miss. So Option B exposes the miss information without forcing the user to consult it. Better than today's `|| 0` but weaker than A.

**Why I'd pick A**:
- Forces the miss case at authoring time (Option B exposes it but doesn't force). Closes the VE/SAP-DM class of bug completely.
- One-line reads vs two-line ceremony. Matters most for the users writing the code — Andreas-class industrial engineers — not for us.
- Falsy-safe by construction (same as B).
- LLM priors match Python/lodash — AI agents writing configs guess the right shape on first try.

**Preempting "just let devs check for undefined"**: the natural objection is that developers could handle it themselves with `typeof v === "undefined"` or `|| default`. True in theory. In practice two things go wrong. `|| default` scattered across every cache read is noise that hurts readability. And users writing `nodered_js` are thinking about their alarm rule or their counter logic, not about JS type coercion — we saw this repeatedly in VE on the SAP DM templates, same miss every time. Force the thinking, don't hope for it.

**Why you might push back toward B**:
- Native JS `Map.get` returns undefined, takes no default. A is a semantic departure; B preserves "it's just a get" optics.
- Go interface coherence argument: B is the faithful mirror of `Get(ctx, key) (any, bool)`; A throws away that shape.
- If you're already leaning B, the "wrong default-name wins the war" argument doesn't apply (only one method exists).

Rewrite all four examples in `javascript-api.md` plus Example 7 / Example 8 in the Go config spec to match whichever shape we pick. For A, the Counter / Previous-value / History / Alarm examples all collapse to one line; the `typeof prev !== "undefined"` contortion becomes `prev === null` against an explicit sentinel.

## The rest

See inline comments on the alarm example (two places, same fix) and on `javascript-api.md` for lifecycle language, missing cycle-time example, missing `return msg;`, and the memory-warning / cache-scope rewrite.

## Follow-up ticket

Please file a child ticket under ENG-3463 for the memory safeguard: threshold, eviction policy, per-processor-vs-global scope. Add it to the Child Tickets table as a Stage 1 gate so it lands before the first release. I'll drop scope guidance on ENG-3463 separately.

docs/processing/javascript-api.md
led0nk: Fixed — all alarm examples now use `msg.payload.value` in both `javascript-api.md` and the Go config spec.

docs/processing/javascript-api.md
led0nk: Fixed.

nodered_js_plugin/nodered_js_plugin.go
led0nk: Fixed in both lines.

docs/processing/javascript-api.md
led0nk: Reworded: "Persists across all messages for the lifetime of the Benthos process. In-memory only, lost on restart."

docs/processing/javascript-api.md
led0nk: Added `return msg;`.

docs/processing/javascript-api.md
led0nk: Added cycle-time example using the two-arg `cache.get(key, default)` pattern.

docs/processing/javascript-api.md
led0nk: Both addressed:
1. Memory warning now calls out the OOM risk explicitly and references ENG-4829 (memory safeguard ticket, Stage 1 gate).
2. Added cache scope note: "the cache is shared across all stages (defaults, conditions, advancedProcessing). A value set in defaults is visible in advancedProcessing within the same message."

led0nk:
All blockers and inline comments from the re-review are addressed:

**Blocker 1: Strip cache config** — Removed the `cache` ObjectField from `NodeREDJSConfigSpec`. No more `cache.backend` or `cache.expiration` in YAML. Both processors hardcode `NewMemoryStore(0)`. Configuration section removed from `javascript-api.md`.

**Blocker 2: SUMMARY.md** — Entry added under Processing.

**Blocker 4: Mandatory default on `cache.get`** — Implemented Option A. `cache.get("counter", 0)` works, `cache.get("counter")` throws TypeError. All examples updated across Go config spec, docs, tests, and tmp yaml files.

**Inline fixes:**
- `msg.payload.value` in all alarm examples (docs + Go config)
- Lifecycle wording: "Persists across all messages for the lifetime of the Benthos process. In-memory only, lost on restart."
- Missing `return msg;` in history example
- Cycle-time example added
- Memory warning strengthened, references ENG-4829
- Cache scope in `tag_processor` documented (shared across all stages)
- CHANGELOG updated for `cache.get(key, default)` signature

**Follow-up ticket:** ENG-4829 filed for memory safeguard (threshold, eviction, scope) as Stage 1 gate under ENG-3463.

krgrsebastian:
I dont like the default case for cache get, this way i dont know (bool for example) if it came from cache or not.

JeremyTheocharis:
**Design resolution trail (for future maintainers):**

Original shape was `cache.get(key) → undefined on miss`. Docs recommended `|| 0` as a fallback, which silently coerced stored `0` / `false` / `""` values. That's the bug class that bit the SAP DM write-flow templates in production and motivated the ticket.

First iteration (Apr 21): mandatory default argument, `cache.get(k, default)`. Forces authoring-time handling, single-line read. @krgrsebastian flagged (Apr 22) that a mandatory default conflates stored values with defaults. A stored `false` then reads the same as a missing key, so you can't tell the two apart.

Final shape (aligned verbally 2026-04-23, pushed 2026-04-24): `cache.get(key)` on miss logs `level=error` via `u.logger.Errorf` and returns `undefined`. A separate `cache.exists(key) bool` helper covers the hit-vs-miss distinction for callers that need it.

Fail-loud happens at the umh-core DFC level, not in the JS runtime. The `level=error` log trips `IsLogsFine` in `pkg/service/benthos/benthos.go`, which moves the Benthos FSM to Degraded, which moves the DFC FSM to Degraded, which causes the deploy-probe's 30s timeout to fire `AtomicEditDataflowcomponent(oldConfig)` (or `AtomicDeleteDataflowcomponent` on fresh deploys). Broken cache flows can't stick. Post-deploy misses surface as a red DFC with an actionable error message pointing at `cache.exists`.

Enforcement lives in umh-core's existing deploy-probe. The JS runtime doesn't need its own gate.

JeremyTheocharis:
Stage 2 question — before we move to the persistent backend (ENG-4752), what's your thinking on how we test whether the current API (if-block + `cache.exists`) is actually working for users, or whether it needs changes?

Drop your take on ENG-3463 whenever makes sense — want the rollout-stage thinking to live next to the ticket you own.

docs/processing/javascript-api.md
```diff
@@ -0,0 +1,126 @@
...
+count++;
+cache.set("count", count);
+msg.payload = count;
+return msg;
+```
+
+### Previous value comparison
+
+```javascript
+var prev = cache.exists("last_value") ? cache.get("last_value") : null;
```
JeremyTheocharis: Nit (doc consistency, not blocking): the nodered_js audience is largely PLC-origin (structured text, ladder logic). They read `IF ... THEN ... ELSE` fluently but find the ternary `? :` cryptic. Suggest rewriting all ternaries in the docs into if-block form so users copy-pasting from docs land on a pattern that reads naturally to them.

nodered_js_plugin/nodered_js_plugin_test.go
```diff
@@ -847,6 +848,277 @@ nodered_js:
...
+`)
+			defer cancel()
+
+			err := handler(context.Background(), newMsg("ignored"))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() int { return len(*msgs) }).Should(Equal(1))
+			Expect(payloadFloat(*msgs, 0)).To(Equal(42.5))
+		})
+
+		It("is safe under concurrent message processing", func() {
```
JeremyTheocharis, line 995: Non-blocking, could be a follow-up. The `cache.get`-on-miss path relies on `u.logger.Errorf` to surface as `level=error` in Benthos stdout. umh-core's `IsLogsFine` parses that to mark the DFC Degraded and fire deploy-probe rollback. If someone later switches to `logger.Warnf` to reduce noise, the gate silently breaks because `IsLogsFine` only treats specific warning substrings as degraded. Worth pinning with one assertion that `level=error` fires on missing-key read. Not a blocker for this PR.

JeremyTheocharis (approved):
Design resolution + Stage 2 question posted above. Two nits inline (nit 1 = doc consistency, fix before merge; nit 2 = test contract pin, follow-up is fine). LGTM once nit 1 is addressed.

docs/processing/javascript-api.md
led0nk: adjusted that

FalconTube (approved):
Looks good now!

PR #331, opened 2026-04-30
docs: prepare release changelog

led0nk:


JeremyTheocharis (commented):
what happened here? why did this go into the 12.3 release? how was it possible to even merge it in there?

in umh-core and managementconsole we habe this explicit check that blocks if you try to modify already released releases to prevent issues like this.

we should probably do it here in betnhos as well

JeremyTheocharis (approved):

led0nk:
documented in eng-4889

PR #333, opened 2026-04-30
fix: remove container scan for no deps

led0nk:
# Description

The fossa-container scan gets removed as there are literally no dependencies or anything that could be scanned here in benthos-umh. Fossa already complaining about exactly this, as there are no existing licenses that could be found.

reference [here](https://github.com/united-manufacturing-hub/benthos-umh/actions/runs/25166148370/job/73772856544?pr=332)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,38 +76,6 @@ jobs:
           provenance: false
           file: ./Dockerfile
 
-  fossa-container:
-    name: FOSSA Container Scan
-    needs: build-docker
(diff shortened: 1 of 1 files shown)


JeremyTheocharis (commented):
maybe keep this idea for umh-core, so when we move it there? so create a ticket / idea usign the code here?

FalconTube (approved):
Makes sense!

led0nk:
> maybe keep this idea for umh-core, so when we move it there? so create a ticket / idea usign the code here?

documented in ENG-4888

PR #334, opened 2026-05-04
chore(ci): update to main, remove on push, label autoPRs

led0nk:
# Description

This PR removes the "on push" Ci runs, which are not needed for any relevant usecase.
Those tests already run in the original PR e.g. against `staging` and afterwards with PLCs in `merge_group`. 
Also PLC tests run whenever there is a `pull_request` targetting `main`. 

Secondly we add the `skip-changelog-guard` label, as this lead to multiple failures of CI within ManagementConsole.

### DIFF
--- a/.github/workflows/update-umh-core-version.yml
+++ b/.github/workflows/update-umh-core-version.yml
@@ -35,8 +35,6 @@ jobs:
           app-id: ${{ secrets.PR_APP_ID }}
           private-key: ${{ secrets.PR_APP_PK }}
           repositories: united-manufacturing-hub
-          labels: |
-            skip-changelog-guard
 
       - name: Checkout united-manufacturing-hub Repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -78,7 +76,6 @@ jobs:
           labels: |
             auto-generated
             dependencies
-            skip-changelog-guard
           draft: false
 
   update-management-console:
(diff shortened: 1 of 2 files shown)


FalconTube (approved):
Good to go after `master` has been renamed to `main`

.github/workflows/update-umh-core-version.yml
```diff
@@ -35,6 +35,8 @@ jobs:
           app-id: ${{ secrets.PR_APP_ID }}
           private-key: ${{ secrets.PR_APP_PK }}
           repositories: united-manufacturing-hub
+          labels: |
+            skip-changelog-guard
```
JeremyTheocharis: are you sure? this will always introduce changelgo changes no?

.github/workflows/update-umh-core-version.yml
```diff
@@ -76,6 +78,7 @@ jobs:
           labels: |
             auto-generated
             dependencies
+            skip-changelog-guard
```
JeremyTheocharis: same here

.github/workflows/update-umh-core-version.yml
led0nk: for this one yes, this is true and landed up in here wrongly

.github/workflows/update-umh-core-version.yml
led0nk: no for MC its only the schema update which will not be relevant in the changelog as introduction of new features will always be part of umh-core

led0nk: but i saw i did the wrong lines, will be corrected

JeremyTheocharis (approved):

PR #335, opened 2026-05-06
fix(opcua): close browse-routine succesfully and handle errors

led0nk:
# Description

While browsing there were scenarios where errors could happen, e.g. `StatusBadNodeID` (could be a misconfigured NodeID). This lead to benthos-umh trying to kill the browse routine, but it wasn't able to succesfully close. Therefore we introduce here a "better" error handling as well as a correct closure of the browsing.
This then leads to a full reconnect as it should. 

Fixes ENG-4195

### Outlook

It might make sense to introduce some general backoff handling as it might happen that opc-servers get hammered with reconnect-tries if someone defines a wrong nodeID.
-> tracked in ENG-4912

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixes
+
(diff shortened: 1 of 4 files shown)


JeremyTheocharis (approved):
LGTM

---
Reviewed in CodeRabbit

PR #340, opened 2026-05-15
fix(modbus): modbus resync after tid mismatches

led0nk:
# Description

The `modbus_plugin` used to drift permanently out of sync when a device replied to a request after that request had already timed out, since the late response would sit in the socket buffer and every following read picked up the response for the previous transaction ID.

The Fix for this was pretty simple as the underlying `grid-x` library already supports this resync, therefore we just had to pass down the timeouts for whenever they're not zero. 

The `SplitHostPort` could ve dropped completely as it was afterwards recomposed again to the same pattern, which would break for IPv6 addresses. If anything goes wrong the underlying `net.Dial` call will anyway detect a potential pattern mismatch of IP & Port.

Fixes ENG-4549

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - OPC-UA input could get stuck while browsing when a configured NodeID did not exist on the server, requiring a manual restart. Browse failures now trigger a clean reconnect
 - OPC-UA input no longer spams `Variant is nil` errors when a node sends a status update without a value. These are harmless and now logged at debug level with the NodeID and status code
 - Modbus TCP input now reconnects immediately on any transport-level error (timeouts, resets, network failures), not just broken pipes. Previously these stuck the connection for up to 10 seconds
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixes
+
+- Modbus TCP input now recovers automatically from transaction-ID mismatches instead of staying permanently out of sync. Affected connections previously had to be restarted manually
```
JeremyTheocharis: `from transaction-ID mismatches instead of staying permanently out of sync`
--> Can you describe instead what the previous user facing behaviour was?  something likle "response transaction id '2767' does not match request '2768'" -->

JeremyTheocharis (approved):

CHANGELOG.md
led0nk: now included the previous user-facing behaviour without stating an exact ID as this changes on each req/resp

PR #345, opened 2026-05-22
ENG-5002: rebirth Sparkplug B input when DATA references uncached aliases

Aaron99B:
## Background

Sparkplug B edge nodes use compact aliases instead of full metric names in DATA messages to save bandwidth on the wire. The bridge resolves these aliases via BIRTH messages the edge publishes when it comes online. If the bridge restarts and the broker holds no retained `NBIRTH`/`DBIRTH` (Ignition's default with MQTT Transmission), the in-memory alias cache is empty but the edge keeps emitting alias-only DATA. Topic Browser shows tags as `…/_historian/alias_<n>` indefinitely.

## Bug

[ENG-5002](https://linear.app/united-manufacturing-hub/issue/ENG-5002): after a bridge restart with no retained BIRTH on the broker, neither `secondary_active` nor `primary` roles request recovery. They silently consume DATA they can't resolve until someone nudges Ignition manually. The Sparkplug v3.0 spec lists "received DATA referencing unknown alias" as a recommended rebirth trigger; the implementation wasn't honoring it.

## Fix

1. **Recovery on unresolved aliases.** `secondary_active` and `primary` roles now request a node-level NCMD/Rebirth when DATA arrives with aliases not in cache. Runs unconditionally (not gated by `request_birth_on_connect`, which controls only proactive discovery).
2. **Unified rebirth dispatch.** Discovery, sequence-gap, and unresolved-aliases rebirths now share one `sendRebirthRequest(key, reason)` function with a single throttle map. Previously the sequence-gap path was unthrottled and could storm the broker on flaky edges.
3. **`request_birth_on_connect` default flip: `false` → `true`.** Bridges proactively recover tag context on connect by default. Set it back to `false` to keep the prior behavior. Ignored under `secondary_passive` (passive bridges never publish commands).
4. **Suppressed-rebirth visibility.** First-time-per-storm suppressions log at `info`; same-dispatch co-fires log at `debug`. Throttling itself is an implementation choice (not Sparkplug B spec behavior); set `birth_request_throttle: 0` to disable.

## Test plan

- [x] `make test-sparkplug` — 254/254 specs pass (10 new specs covering positive discovery, throttle disable, same-dispatch co-fire, role-gate coverage for all 3 reasons)
- [x] `make lint`, `make fmt`, `go vet` clean
- [x] `go build ./sparkplug_plugin/...` clean (verifies test helpers no longer ship in production binary)
- [x] **Manual end-to-end** against real Ignition (matches the issue's reproduction):
  - UMH-core + ManagementConsole running locally
  - Ignition publishing Sparkplug B with aliases enabled in MQTT Transmission
  - Sparkplug input bridge configured with `role: secondary_active`
  - In ManagementConsole, disable the bridge → wait ~20s → re-enable
  - Run `mosquitto_sub -h <broker> -t 'spBv1.0/+/NCMD/+' -v` in another terminal
  - **Expected**: an NCMD message appears within a second of the first incoming DDATA/NDATA; Topic Browser tags resolve to real metric names after Ignition's next BIRTH cycle

## Known follow-ups

- Integration test asserting the actual NCMD lands on a real broker (the unit suite asserts the recovery path executes via the `alias_rebirths` counter, but the publish itself is unreachable without an MQTT fixture)
- TTL eviction for the `birthRequested` throttle map; currently unbounded by unique node IDs ever observed
- `processBirthMessage` does not clear the throttle entry when NBIRTH actually arrives; a second protocol violation on the next NDATA is throttled even though BIRTH has reset protocol state
- ENG-5003 (NBIRTH/DBIRTH published with `retain=true`) — separate ticket on the output side

## Changelog Entry

**Component:** benthos-umh (entries in benthos-umh's own `CHANGELOG.md`)
**Category:** Fixes

- Sparkplug B input: bridges now request a rebirth when DATA references aliases the cache hasn't seen (typically after a bridge restart with no retained `NBIRTH`/`DBIRTH` on the broker). Previously tags surfaced as `…/_historian/alias_<n>` until something external triggered recovery. Related, `request_birth_on_connect` now defaults to `true`, so `secondary_active`/`primary` bridges also proactively rebirth newly seen nodes on connect; set it to `false` to keep the prior behavior (ignored under `secondary_passive`).


sparkplug_plugin/sparkplug_b_input.go
```diff
@@ -1362,6 +1394,39 @@ func (s *sparkplugInput) requestBirthIfNeeded(deviceKey string) {
...
+// bridge missed the corresponding NBIRTH/DBIRTH (e.g. broker held no retained BIRTH at the
+// time this bridge connected).
+//
+// Throttled via the same birthRequested map and BirthRequestThrottle window as the
+// discovery-rebirth path. Sharing the map (rather than a separate one) prevents the two code
+// paths from racing on the same node and flooding the broker. Role suppression for
+// secondary_passive is delegated to sendRebirthRequest, but we short-circuit here as well so
+// passive bridges never even touch the throttle bookkeeping.
+func (s *sparkplugInput) requestRebirthForUnresolvedAliases(nodeKey string, unresolvedCount int) {
+	if s.config.Role == RoleSecondaryPassive {
```
JeremyTheocharis: maybe make it the other way around: only do this if secondary active or primary (so default off)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
...
 - Sparkplug B input: field descriptions and the primary-role startup log now state that `identity.edge_node_id` is used as the Sparkplug v3.0-compatible `host_id` in the STATE topic (`spBv1.0/STATE/<host_id>`).
+- Sparkplug B input: `secondary_active` and `primary` roles now publish `NCMD/Node Control/Rebirth` when DATA arrives with aliases that aren't in the in-memory cache — typically after a bridge restart where the broker has no retained `NBIRTH`/`DBIRTH`. Previously the cache stayed empty and tags surfaced in the Topic Browser as `…/_historian/alias_<n>` until something external (e.g. an upstream rebirth from an HMI or commissioning tool) triggered recovery. This recovery is unconditional for non-passive roles and is *not* gated by `request_birth_on_connect` (which only affects proactive discovery rebirths on new-node observation). Throttled by `birth_request_throttle` (default 1s) to prevent storms; suppressed for `secondary_passive` like the existing rebirth paths. New metric `alias_rebirths` counts triggered recoveries; new metric `alias_resolution_failures` counts per-metric cache misses (previously folded into the generic `messages_errored`).
```
JeremyTheocharis: this changelog is rather complex. Is it possible to make it simpler? Previously secondary active as well as primary were not sending rebirth request messages when encountering aliases. This is now fixed. 

or something

JeremyTheocharis (changes requested):
<sub>Requested changes in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/united-manufacturing-hub/benthos-umh/pull/345)</sub>

sparkplug_plugin/sparkplug_b_input.go
Aaron99B: Changed that

CHANGELOG.md
Aaron99B: Made it shorter and simpler

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 
 - Sparkplug B input: `identity.group_id` now filters the MQTT subscription by default. Previously an empty `subscription.groups` caused the plugin to subscribe to every Sparkplug group on the broker (`spBv1.0/+/#`) regardless of `identity.group_id`. To restore the old behavior, set `subscription.groups: ["+"]` explicitly.
 - Sparkplug B input: field descriptions and the primary-role startup log now state that `identity.edge_node_id` is used as the Sparkplug v3.0-compatible `host_id` in the STATE topic (`spBv1.0/STATE/<host_id>`).
+- Sparkplug B input: `secondary_active` and `primary` roles now send a rebirth request when DATA arrives referencing aliases the bridge hasn't seen (typically after a bridge restart with no retained `NBIRTH`/`DBIRTH` on the broker). Previously these roles stayed silent and tags surfaced as `…/_historian/alias_<n>` until something external triggered recovery.
```
led0nk: correct me if i am wrong, but in my opinion this fix here is 1 or maximum 2 changelog-entries regarding the flip of the flag `request_birth_on_connect` as well as the fix for unresolved aliases

Also i'd love to have this a bit shorter here, maybe it works better with the new writing skill of jeremy?

led0nk (changes requested):
As we already discussed this approach works for me, but i would like to keep the changelog precise. Afterwrds this is ready to go from my pserpective

CHANGELOG.md
Aaron99B: Changed that

led0nk (approved):
good for now, in the future i would make the changelog even less verbose

JeremyTheocharis (approved):

PR #347, opened 2026-05-26
ENG-5003: Sparkplug B output retain=false for Edge Node messages

Aaron99B:
## Summary
- Sparkplug B v3.0 §5.4 requires all non-STATE messages (NBIRTH, DBIRTH, NDEATH, DDEATH, LWT) to use `retain=false`; only Primary Host STATE may be retained. The `edge_node` output plugin previously retained all five — this PR fixes that.
- Late subscribers will no longer receive stale alias-to-name maps from the broker's retained store; they learn node state via `\$sparkplug/certificates/#` on a Sparkplug-Aware broker, or by requesting rebirth via `NCMD/Node Control/Rebirth`.
- Wire-visible behavior change is called out in the CHANGELOG.

## Linear
https://linear.app/united-manufacturing-hub/issue/ENG-5003

## Scope note
The Linear ticket originally scoped only NBIRTH/DBIRTH (the user-authored description claimed NDEATH retain=true was "correct per spec"). Self-review verified against the Eclipse Sparkplug normative statements that *all* non-STATE messages MUST be retain=false. This PR expanded scope to also flip NDEATH/DDEATH (line 766) and the LWT registration (line 442).

## Test plan
- [x] `make test-sparkplug` — 241/241 specs pass (4 new ENG-5003 specs covering NBIRTH / DBIRTH / NDEATH / DDEATH retain assertions; new Ginkgo unit test uses a hand-rolled paho \`mqtt.Client\` fake)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
...
 - Sparkplug B input: field descriptions and the primary-role startup log now state that `identity.edge_node_id` is used as the Sparkplug v3.0-compatible `host_id` in the STATE topic (`spBv1.0/STATE/<host_id>`).
+- Sparkplug B output (`role: edge_node`): NBIRTH, DBIRTH, NDEATH/DDEATH (graceful Close) and the LWT registration were all published with `retain=true`, violating Sparkplug B v3.0 §5.4 which requires all non-STATE messages to be published with `retain=false` (only Primary Host STATE messages may be retained). Late subscribers received stale alias-to-name maps from the broker's retained store, producing subtly wrong name resolution after tag configuration changed. All five publishes now use `retain=false`. Wire-visible behavior change: late subscribers connecting after the Edge Node will no longer receive a retained BIRTH/DEATH from the broker — they must subscribe to `$sparkplug/certificates/#` on a Sparkplug-Aware broker, or request rebirth via `NCMD/Node Control/Rebirth`. Operators running `CleanSession=false` clients should re-subscribe to BIRTH topics from a Primary Host context.
```
JeremyTheocharis: the changelog is rather complex, i think this can be described user facing using a single sentence.

JeremyTheocharis (commented):
setting it to retained was a deliberate decision to ensure that secondary_passive always has access to at least some messages. but it is not sparkplugb compliant.

fine either way

CHANGELOG.md
Aaron99B: I simplified the changelog

JeremyTheocharis (commented):
Reviewed against the new `writing-standards:changelog-writing` skill. Two issues with the current bullet:

1. **Frames the bug, not the fix.** Per Rule 11 + the Trust Test, Fixes should describe the corrected behavior. Leading with "were all published with `retain=true`, violating §5.4" reads as a confession of a spec violation rather than the shipped fix.
2. **Buried breaking change.** Per "Breaking changes go inline → state the required user action", the consumer-side action (request rebirth via `NCMD/Node Control/Rebirth`) belongs as a clearly named action, not as "must request a rebirth to receive it" at the tail. Not severe enough for an `**Action required before upgrade:**` callout (no startup failure), but downstream consumers relying on retained BIRTHs will silently miss them.

Minor:
- "NDEATH/DDEATH (graceful Close)" — "graceful Close" is internal Go-method naming; readers don't need it.
- Reframe the spec citation as compliance, not violation (trust-safe framing).

Suggestion below leads with the corrected behavior, names the downstream action explicitly, and keeps the spec reference for Sparkplug-aware readers.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Sparkplug B input: `identity.group_id` now filters the MQTT subscription by default. Previously an empty `subscription.groups` caused the plugin to subscribe to every Sparkplug group on the broker (`spBv1.0/+/#`) regardless of `identity.group_id`. To restore the old behavior, set `subscription.groups: ["+"]` explicitly.
 - Sparkplug B input: field descriptions and the primary-role startup log now state that `identity.edge_node_id` is used as the Sparkplug v3.0-compatible `host_id` in the STATE topic (`spBv1.0/STATE/<host_id>`).
+- Sparkplug B output (`role: edge_node`): NBIRTH, DBIRTH, NDEATH/DDEATH (graceful Close) and the LWT registration were all published with `retain=true`, violating Sparkplug B v3.0 §5.4 which requires all non-STATE messages to be published with `retain=false`. Wire-visible behavior change: late subscribers connecting after the Edge Node will no longer receive a retained BIRTH/DEATH from the broker and must request a rebirth to receive it.
```
JeremyTheocharis:
```suggestion
- Sparkplug B output (`role: edge_node`): NBIRTH, DBIRTH, NDEATH, DDEATH, and the LWT registration are now published with `retain=false`, matching Sparkplug v3.0 §5.4. Late subscribers no longer receive retained BIRTHs from the broker; they must request one via `NCMD/Node Control/Rebirth` or rely on a Sparkplug-Aware broker's `$sparkplug/certificates/#` topics.
```

Aaron99B:
Parked until more data regarding Sparkplug B output is gained

PR #349, opened 2026-05-27
chore: add pr-comment for image tag

led0nk:
# Description

This PR introduces PR-comments for docker-image tags. It should improve testing and development time for devs, which need to interact with benthos-images.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,6 @@ jobs:
       pkg: ${{ steps.filter.outputs.pkg }}
       cmd: ${{ steps.filter.outputs.cmd }}
       gomod: ${{ steps.filter.outputs.gomod }}
-      code: ${{ steps.filter.outputs.code }}
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -105,10 +104,6 @@ jobs:
               - 'cmd/**'
             gomod:
               - 'go.{mod,sum}'
-            code:
-              - '**/*.go'
-              - 'go.{mod,sum}'
-              - 'Dockerfile'
   fossa:
     name: FOSSA License Scan
     needs: detect-changes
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -312,6 +317,45 @@ jobs:
     uses: ./.github/workflows/test-classic-to-core.yml
     secrets: inherit
 
+  comment-pr:
+    name: Comment PR with Image Tag
+    needs: detect-changes
+    if: github.event_name == 'pull_request' && needs.detect-changes.outputs.code == 'true'
+    runs-on: ubuntu-latest
```
FalconTube: Was using the `github ubuntu` runner intentional or do you want to use `depot` instead?
This step does not really do anything computationally intensive, but we also still have plenty of depot minutes left and might be good to have everything at one place.
I don't have a strong opinion tho.

.github/workflows/ci.yml
```diff
@@ -312,6 +317,45 @@ jobs:
...
+      pull-requests: write
+    steps:
+      - name: Wait for Docker image build
+        uses: lewagon/wait-on-check-action@9312864dfbc9fd208e9c0417843430751c042800 # v1.7.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          check-name: build-docker
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          wait-interval: 20
+          allowed-conclusions: success
```
FalconTube: Ah, I see why you need this action here.
But it looks like quite a hack to me.
It might make sense to adjust the `on: ...` behavior in `main.yml` instead and make our lives simpler :)
Look at the general comment that I left for more info

FalconTube (changes requested):
While looking at this, I realized that we are using quite a few hacks in `main.yml` and therefore now also in `ci.yml`.

It might make sense to change the `on: ...` behavior in `main.yml` to the following:

```yaml
on:
  push:
    branches: [master]
    tags: [v*]
  pull_request:
```

That way you can get rid of a lot of work in the `comment-pr` step in `ci.yml` and just move the `comment-pr` to `main.yml`, because it would now also run on PR.

From what I can see, the only thing we then miss is that we don't get a docker image `on push`, when there is not open `PR` for it. But imo we don't need that, right?

Am I missing something? Let's maybe talk about this quickly tomorrow :)

---

Here some Klaus summary on what it `hacky`. Probably can adjust a few of them, but maybe not all

```md
  ┌─────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────┐
  │                                    Hack                                     │                            Why it exists                            │
  ├─────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ build-binaries: if: startsWith(github.ref, 'refs/tags/')                    │ trigger too broad — fires on every branch push                      │
  ├─────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ release: if: needs.build-binaries.result == 'success'                       │ defensive guard because build-binaries gets skipped on non-tag runs │
  ├─────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ update-umh-core: if: needs.release.result == 'success'                      │ same cascading guard                                                │
  ├─────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ ref: ${{ github.event.pull_request && github.head_ref || github.ref_name }} │ PR context never exists on push event — dead ternary                │
  ├─────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ type=ref,event=pr in docker meta                                            │ never fires since there's no pull_request trigger — dead config     │
  ├─────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ wait-on-check-action polling in ci.yml                                      │ cross-workflow wait because comment-pr can't use needs:             │
  └─────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────┘

```

.github/workflows/ci.yml
led0nk: this was intentional, just took a look at umh-core and realized that it might be simpler & much faster to use the `github ubuntu` one here as this is a really quick action

.github/workflows/ci.yml
led0nk: you're absolutely roight, adjusted this in main.yml to run on `pull_request` this makes it easier to include the comment here without workarounds

.github/workflows/main.yml
```diff
@@ -76,6 +73,34 @@ jobs:
...
           file: ./Dockerfile
 
+  comment-pr:
+    name: Comment PR with Image Tag
+    needs: build-docker
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
```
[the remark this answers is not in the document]
led0nk: not a relevant issue for us

.github/workflows/release.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: artifact-${{ matrix.goos }}-${{ matrix.goarch }}-${{ github.run_id }}-${{ github.run_attempt }}
+          path: ./dist/${{ env.OUTPUT_NAME }}
+
+  release:
+    needs: build-binaries
+    if: needs.build-binaries.result == 'success'
```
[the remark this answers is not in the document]
led0nk: we don't really need this for a release currently

FalconTube (approved):
Nice, looks great now!

PR #355, opened 2026-06-02
Feat: open protocol 

krgrsebastian:


krgrsebastian:
## Field test against a real controller (Atlas Copco PF6000-family, TRUMPF production line)

We pointed the PR image (`sha-707e94a`) at a real tightening controller (VHF APL5 tool on a PF6000-family virtual station) plus the art-x emulator locally. Analysis only — no code changes yet. Summary of findings:

### What works

Login (MID 0001), subscribe (MID 0060), MID 0005 confirmation, keep-alives, and the 18-tag MID 0061 fan-out all work — verified against both the art-x emulator and the real controller.

### 1. `subscribe` config trap: data MID vs. subscription MID

A first-time user naturally writes `subscribe: [61]` ("I want MID 0061 data"), but the field takes the subscription request MID (`60`). The result is a `Connect` failure loop:

```
Failed to connect to open_protocol: open protocol command error: MID 0061 failed with error code 97
```

(The real controller NAKs with error 97; the art-x emulator with 99 — either way an opaque retry loop.)

Since valid subscription MIDs are even (data = MID+1, ack = MID+2), a cheap config-time validation could turn this into an immediate, self-explanatory startup error, e.g. *"subscribe takes the subscription MID (60), not the data MID (61)"*.

### 2. No graceful logout (MID 0003)

`Session.Close()` and all error paths drop TCP without sending MID 0003 (communication stop). The controller always sees a dirty disconnect. Harmless against the emulator; against a production controller it interacts badly with finding 3 and 4.

### 3. No reconnect backoff

While misconfigured (finding 1), the input retried `Connect` roughly once per second indefinitely. Against a real controller that's a login + dirty disconnect every second, in production. Some backoff (or at least a cap/jitter) would be kinder to shop-floor hardware.

### 4. Real-controller interaction: tool lock policy ("Disconnect: Lock tool")

PF6000 virtual stations can be configured so the tool is **locked whenever no Open Protocol client is connected** (the SQS3 line-control pattern). On the test line this produced a "Tool locked by open protocol" event (code 4011) on the controller HMI as soon as Open Protocol was enabled on the virtual station, before our client ever connected.

Two implications for this input:

- In the Atlas Copco reference integration, the client is expected to send **MID 0043 (enable tool)** after connecting; connecting alone does not necessarily release the lock. This input intentionally sends no tool-control MIDs (verified: only 0001, subscribe MIDs, acks, 9999), so on such stations the tool stays locked even with a healthy session.
- Worth a docs note at minimum: for pure monitoring, the virtual station's Open Protocol *Disconnect* setting should not be "Lock tool"; for line-control use, the input would need an opt-in like `enable_tool_on_connect` (MID 0043) — deliberately out of scope for now since it makes data collection production-critical.

Happy to turn 1–3 into small follow-up commits on this branch if we agree on direction.

PR #356, opened 2026-06-03
feat: generate mapping for read/write-flows

led0nk:
# Description

For the introduction of Write-Flows we need some mapping for the Read <-> Write - flow-dependency. In specific we only want to connect to 1 device, which refers to the corresponding read & write flow protocols to be the same. 
E.g. if you use a Read-Flow to read data from `opcua` (which is an input-protocol), you would also need to use `opcua` (output-protocol) to connect to the same device.

Therefore we need to introduce some mapping to validate in the UI as well as in `umh-core`. 
We decided to generate this from within `benthos-umh` as this is the first source of truth, when it comes to reading the registered input and output-plugins.

This PR introduces the generation of a `json` file which maps input-protocols (`Read`)  to output-protocols (`Write`). Also for later stages we introduced the `umh`-flag, which shows if this is a self-written one. 
On release this then generates in addition to the `ui.json` and `codemirror.json` the `read-write-mapping.json` and adds it to the PR for ManagementConsole.

Also it uses some overrides (`readWriteOverrides`) , which gives you the ability to define certain pairs of `Read` and `Write` which are intentionally not matched by the generator.

As this is not user-facing by any means there is no changelog-entry.

Fixes ENG-5033


cmd/schema-export/main.go
```diff
@@ -58,6 +60,22 @@ func main() {
...
+			os.Exit(1)
+		}
+		err = os.WriteFile("read-write-mapping.json", mappingData, 0o644)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error writing output: %v\n", err)
+			os.Exit(1)
+		}
+		fmt.Printf("✅ Generated read→write mapping to read-write-mapping.json (%d pairs)\n", len(mapping))
+		return
+	}
```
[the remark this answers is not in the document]
led0nk: included

cmd/schema-export/read_write_map.go
```diff
@@ -0,0 +1,131 @@
...
+		Read:  "aws_dynamodb_cdc",
+		Write: "aws_dynamodb",
+	},
+}
+
+// buildMapping derives the read->write pairing from all registered inputs and outputs, applies overrides, and returns it sorted by plugin name.
+func buildMapping(env *service.Environment, overrides []pluginPair) []pluginPair {
+	registeredInputs := make(map[string]bool)
+
+	// get non-deprecated inputs from redpanda-connect and benthos-umh
```
FalconTube: Super duper mega nitpick: Make this declaration formatting the same as for the outputs few lines below

FalconTube (approved):
Looks great! 
Just have one super nit, because I know you care about this stuff ;)

cmd/schema-export/read_write_map.go
led0nk: you're right - don e

PR #363, opened 2026-06-08
fix: make string types directly use json.Marshal conv (eng-5011)

led0nk:
# Description

In `opcua_plugin` when reading from strings it could happen that you get the edge-case of reading large numeric strings, which will then be not correctly converted due to "raw" data extraction.

We now use direct `json.Marshal` for this to work around any edge-cases that could happen e.g. with large numeric strings, `"` in strings or special characters in strings.

Fixes ENG-5011

### DIFF
--- a/opcua_plugin/core_read.go
+++ b/opcua_plugin/core_read.go
@@ -69,9 +69,6 @@ func (g *OPCUAConnection) getBytesFromValue(dataValue *ua.DataValue, nodeDef Nod
 	case float64:
 		b = append(b, []byte(strconv.FormatFloat(v, 'f', -1, 64))...)
 		tagType = "number"
-	case string:
-		b = append(b, []byte(v)...)
-		tagType = "string"
 	case bool:
 		b = append(b, []byte(strconv.FormatBool(v))...)
 		tagType = "bool"
@@ -106,7 +103,7 @@ func (g *OPCUAConnection) getBytesFromValue(dataValue *ua.DataValue, nodeDef Nod
 		b = append(b, []byte(strconv.FormatUint(v, 10))...)
 		tagType = "number"
 	default:
-		// Convert unknown types to JSON
+		// Convert unknown and string types to JSON
 		jsonBytes, err := json.Marshal(v)
 		if err != nil {
 			g.Log.Errorf("Error marshaling to JSON: %v", err)
--- a/opcua_plugin/core_read.go
+++ b/opcua_plugin/core_read.go
@@ -103,7 +103,8 @@ func (g *OPCUAConnection) getBytesFromValue(dataValue *ua.DataValue, nodeDef Nod
 		b = append(b, []byte(strconv.FormatUint(v, 10))...)
 		tagType = "number"
 	default:
-		// Convert unknown and string types to JSON
+		// Strings must be JSON-quoted too: downstream parses the body as JSON, and raw
+		// bytes would turn numeric-looking strings into lossy numbers.
 		jsonBytes, err := json.Marshal(v)
 		if err != nil {
 			g.Log.Errorf("Error marshaling to JSON: %v", err)
--- a/opcua_plugin/core_read_test.go
+++ b/opcua_plugin/core_read_test.go
@@ -17,10 +17,9 @@ package opcua_plugin
 import (
 	"encoding/json"
 
+	"github.com/gopcua/opcua/ua"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"github.com/gopcua/opcua/ua"
 )
 
 // ENG-5011: an OPC UA String value must be emitted as a valid JSON string so it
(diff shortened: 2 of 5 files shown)


opcua_plugin/core_read_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/gopcua/opcua/ua"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// ENG-5011: an OPC UA String value must be emitted as a valid JSON string so it
```
JeremyTheocharis, line 30: i understand the problem, but i do not udnerstand how this is fixed. are we here simply not emitting isntead of a string a json?

led0nk: this is fixed by json.Marshalling the string (which ends up expected for the processing part as the other message-values) instead of injecting the value as raw bytes which then gets incorrectly parsed as number (if it was a number)

JeremyTheocharis: got it for the opcua layer: the quoted body means the first downstream JSON parse (the `json.Unmarshal` in `nodered_js_plugin.go` `ConvertMessageToJSObject`, which ENG-5011 names as the damage point) now yields a string instead of a float64, so precision survives into the JS stage.

but I want to double-check the end-to-end story: with `datatype` meta unset, tag_processor's `autoConvertValue` still runs `strconv.ParseFloat` on string values (tag_processor_plugin.go, `case string:`) and converts `"83280661010132726"` right back to `8.328...e+16`. so the value only survives the full pipeline when the pipeline also sets `msg.meta.datatype = "string"` (the v0.44.19 override, which this PR makes effective for numeric-looking string tags for the first time).

two questions:
1. did you verify the full chain (opcua → tag_processor) with `msg.meta.datatype = "string"` set?
2. should we state in the changelog/docs that `msg.meta.datatype = "string"` is still required for numeric-looking string tags, since auto-detection will re-convert them by design?

opcua_plugin/core_read.go
```diff
@@ -106,7 +103,7 @@ func (g *OPCUAConnection) getBytesFromValue(dataValue *ua.DataValue, nodeDef Nod
 		b = append(b, []byte(strconv.FormatUint(v, 10))...)
 		tagType = "number"
 	default:
-		// Convert unknown types to JSON
+		// Convert unknown and string types to JSON
```
JeremyTheocharis: can we capture the "why" here? the comment says what the code does, but the next reader will ask the same question I did in the test file.

```suggestion
		// Strings deliberately take this path too (ENG-5011): the message body
		// is parsed as JSON downstream, so a string must be emitted as a quoted
		// JSON string literal. Raw unquoted bytes would make a numeric-looking
		// string like "83275631010238526" parse as a number and lose precision.
```

opcua_plugin/core_read_test.go
```diff
@@ -0,0 +1,49 @@
...
+// round-trips as a string downstream. A numeric-looking string ("83280661010132726")
+// must not be re-parsed as a (lossy) number by the downstream json.Unmarshal.
+var _ = Describe("getBytesFromValue string encoding", func() {
```
JeremyTheocharis, line 33: this pins `getBytesFromValue`, which is good, but the contract that actually broke is one level up: the benthos message body has to come back as the original string from a structured parse (that is what `nodered_js_plugin.ConvertMessageToJSObject` effectively does). can you add one test that goes through `createMessageFromValue` and asserts the round-trip on the message itself? sketch (compiles and passes against your branch):

```go
It("round-trips a string through the benthos message", func() {
	input := &OPCUAInput{OPCUAConnection: &OPCUAConnection{}}
	value := "83280661010132726"
	dv := &ua.DataValue{Value: ua.MustVariant(value), Status: ua.StatusOK}
	nodeDef := NodeDef{NodeID: ua.NewNumericNodeID(0, 1234)}

	msg := input.createMessageFromValue(dv, nodeDef)
	Expect(msg).NotTo(BeNil())

	structured, err := msg.AsStructuredMut()
	Expect(err).NotTo(HaveOccurred())
	Expect(structured).To(Equal(value))
})
```

that also guards how the body gets attached to the message, not just the byte encoding. (the opc-plc integration specs that validate datatypes via `checkDatatypeOfOPCUATag` are currently `Skip`'d due to opcsimv2 misbehavior, and they `continue` on parse errors — so string tags were never reliably type-checked there.)

opcua_plugin/core_read.go
led0nk: updated but without issue-no (don't like this approach)

opcua_plugin/core_read_test.go
led0nk: improved the testing here

JeremyTheocharis (approved):

opcua_plugin/core_read_test.go
led0nk: addressed via documentation, also verified full-chain with msg.meta.datatype being set. Test is also adjusted

PR #365, opened 2026-06-08
ENG-5087: fix Sparkplug B output dropping tag_processor messages

Aaron99B:
## What

A standard `tag_processor → sparkplug_b` flow silently dropped every message — the broker only ever saw an empty `NBIRTH`, never `DBIRTH`/`DDATA`.

## Why

The output read the metric value from a payload field named after the tag, but `tag_processor` emits the value under `value`. Nothing matched, so every message was acked and dropped with no error.

## Fix

- Extract the value and metric name via the same converter the Sparkplug B **input** already uses — value from `value`/`val`/`data`/`measurement`, metric name `virtual_path:tag_name` (or just `tag_name` when `virtual_path` is empty).
- Make `virtual_path` optional.
- Drop non-scalar payloads instead of publishing a stringified map, and warn on dropped messages so a misconfigured pipeline is observable.

## Testing

- New unit tests for the auto-extract path and the construction seam; full sparkplug suite (261 specs) passes; `go vet` and `golangci-lint` clean.
- End-to-end: ran the built binary against a real Modbus device with the original config (no `nodered_js` workaround) — publishes `DBIRTH` + `DDATA` with the correct tag values.

Closes ENG-5087

### DIFF
--- a/sparkplug_plugin/sparkplug_b_output.go
+++ b/sparkplug_plugin/sparkplug_b_output.go
@@ -1251,8 +1251,15 @@ func (s *sparkplugOutput) extractMessageData(msg *service.Message) (map[string]i
 		"autoExtractTagName: %t, configured metrics: %d, structured payload: %+v",
 		s.autoExtractTagName, len(s.metrics), structured)
 
+	// A tag-scoped message (tag_processor flow: tag_name in metadata) represents
+	// exactly one tag; its metric identity comes from the metadata. Once it is
+	// handled here, the static-metric sweep below must be skipped so the single
+	// value does not fan out across every configured metric sharing its value_from.
+	handledViaTagName := false
+
 	if s.autoExtractTagName {
 		if tagName, exists := msg.MetaGet("tag_name"); exists {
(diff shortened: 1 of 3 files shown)


JeremyTheocharis (commented):
@led0nk should review

sparkplug_plugin/sparkplug_b_output.go
```diff
@@ -51,8 +51,8 @@ import (
 	"github.com/united-manufacturing-hub/benthos-umh/sparkplug_plugin/sparkplugb"
 )
 
-func init() {
-	outputSpec := service.NewConfigSpec().
+func sparkplugOutputConfigSpec() *service.ConfigSpec {
```
led0nk: could you do me a favor and make this exactly like the other plugins, meaning the ConfigSpec being stored in a `const` and then called in the init function here?

other than that looks good to me

led0nk (changes requested):

led0nk (approved):

led0nk (approved):

PR #366, opened 2026-06-09
fix: sort required fields alphabetically in schema export

FalconTube:
- Sorts `required` fields of schema in alphabetical order. 
- Minimizes diffs in future revisions of schemata


JeremyTheocharis (approved):
LGTM if @led0nk also approves

led0nk (approved):
works for me, was wondering if its only about the `required` fields

FalconTube:
@led0nk yes, only the required fields produce lots of diffs.
But will open a separate PR about ordering and preserving golang structs soon.
Has different reasons tho :)

PR #367, opened 2026-06-10
feat(uns_plugin): add uns_beta input wrapping the official redpanda engine (ENG-5105)

JeremyTheocharis:
## Background

The `uns` input is a hand-rolled franz-go consumer. It disables autocommit and commits with one shared ack across all in-flight batches, so committing any batch advances the offset past everything polled so far.

## Bug (ENG-5094)

When an output rejects a message and a later message succeeds, the commit advances past the rejected one and it is never redelivered: silent data loss. `uns_plugin/uns_input_nack_commit_repro_test.go` reproduces it against a real in-memory Kafka broker (kfake): NACK message A, ACK message B, and the committed offset jumps past A with zero redeliveries.

## Fix

A new input, `uns_beta`, delegates consuming and committing to the official `redpanda` Connect input (through benthos `OwnedInput`), and adds `umh_topics` key filtering and `umh_topic` / `kafka_msg_key` metadata on top. The delegated input never commits past a message your pipeline rejects: a NACKed message is redelivered until it is accepted, and only then does the offset advance. Filtered records, which are never delivered, still advance the offset so the partition does not stall.

The user-facing template exposes four fields:

- `consumer_group`: required; must be unique to the consumer.
- `umh_topics`: required; a consumer states which topics it wants rather than defaulting to the whole namespace.
- `broker_address`: defaults to the embedded Redpanda (`localhost:9092`).
- `kafka_topic`: defaults to `umh.messages`.

`uns` itself is untouched; it gets repointed once `uns_beta` proves out (ENG-5104 has the rollout plan). User docs are in `docs/input/uns-beta-input.md`.

## Selective-consumer allocation

`uns_beta` is built for consumers that subscribe to a slice of a high-volume namespace. A `key_pattern` pre-filter in connect's record reader skips the message build for non-matching records (`select_one`: ~32 down to under 3 allocs per record), keeping the poll's last record as a no-metadata placeholder so the offset still advances. `uns_beta` applies the same `umh_topics` set again to the delivered batch, and connect's checkpointer-driven commit is unchanged.

connect's reader has no hook for this and the function is unexported, so it ships as a build-time patch, not a fork and not a vendored tree:

- `patches/connect-key-filter.patch` is applied by `make patch-connect` onto a throwaway copy of the pinned module. The `replace` goes to a side modfile (`go.patched.mod`) so the committed `go.mod` never changes.
- Every shipped build (`make build`, Dockerfile, release) patches. `git apply --check` fails the build when a connect bump moves the patched code.
- `uns_beta` tests run in a patched CI lane (`test-uns-patched`); the legacy `uns` tests stay unpatched.
- Commits stay gapless even when an entire poll is filtered out. Pre-filtered records are counted by connect's `key_omitted_records`.
- Long-term, upstream a `key_pattern` field to connect and remove the patch.

## Reviewer attention

Trade-off: maintaining a build-time patch of a dependency, mandatory across builds. Filtering on the native record key also closes the ENG-5125 `kafka_key`-spoof vector at the source.

Modeled as two registered plugins (`uns_beta` template + `uns_beta_reader` input), not one: a managed input that keeps `kafka_lag` and metrics on the outer stream can only come from a framework-parsed `input:` field, so the lean user surface and the wrapping behavior cannot share one registration. It can be collapsed into a single Go input, but that builds the `redpanda` input on a throwaway manager and loses its `kafka_lag` and metrics.


JeremyTheocharis:
Superseded by #395 — the single-plugin form (Option B) was preferred by Aaron and Janik on 2026-06-26; the template form is dropped. The converged branch (`eng-5105-uns-beta-final`) carries the key-filter patch forward, so that work is not lost.

PR #368, opened 2026-06-10
ENG-5126: Sparkplug B input — persist metric datatype from BIRTH, decode signed integers

Aaron99B:
## Summary

Fixes [ENG-5126](https://linear.app/united-manufacturing-hub/issue/ENG-5126/sparkplug-b-input-metric-datatype-dropped-on-ndata-spb-datatype): community report that `spb_datatype` was missing in the Topic Browser on NDATA messages and values were wrong, while alias→name resolution worked.

Per the Sparkplug B spec, NDATA/DDATA carries only `alias + value`; the BIRTH certificate defines `name` and `datatype`, and the host must remember both. The input's `AliasCache` remembered only the name:

1. **Datatype never cached.** On DATA messages `metric.Datatype` stayed nil, so the `spb_datatype` metadata was never set and the UMH conversion path fell back to `"unknown"`.
2. **Datatype never applied.** Sparkplug packs signed integers as two's-complement into the unsigned `int_value`/`long_value` wire fields. Without the datatype, an `Int32` of `-12` surfaced as `4294967284` (wrong even on the BIRTH message itself).

## Changes

- `AliasCache` now stores `alias → {name, datatype}`; `ResolveAliases` restores the datatype onto DATA metrics the same way it already restored the name (a datatype sent on the DATA metric itself wins)
- New `decodeIntValue`/`decodeLongValue` helpers reinterpret the wire value per datatype in both extractors (`extractMetricValue` for the JSON payload, `extractMetricValueRaw` for UMH conversion), so `Int8`/`Int16`/`Int32`/`Int64` decode as signed; UInt*/Float/Double/Bool/String paths unchanged
- Dropped the dead `legacyAliasCache` field (declared and initialized, never read)
- Changelog entry under `## [Unreleased]` → `### Fixes`

## Testing

TDD: all new specs in `sparkplug_b_datatype_test.go` were written first and watched fail (8 failing, 5 control specs passing) before any production code changed.

- Cache-level specs: datatype restored on resolve, publisher-sent datatype not overwritten, datatype-less BIRTH still resolves the name
- NBIRTH→NDATA flow specs through the same path production uses: `spb_datatype` metadata present on NDATA, plus a signed/unsigned decode table (Int8/Int16/Int32/Int64 negative, positive passthrough, UInt32/UInt64 unchanged, Float untouched)
- Full suite: `make test-sparkplug` — 276/276 passing
- Reproduced live before the fix with a minimal Sparkplug publisher (retained NBIRTH `Int32 -7`, alias-only NDATA `-12`) against a umh-core bridge: `spb_datatype` absent, value read `4294967284`

## Notes

- DateTime metrics still pass through as raw epoch-millis — deliberate; converting them is a behavior change tracked in ENG-5126 as out of scope
- Complex types (DataSet/Template/Bytes) remain unhandled (dropped) as before — separate gap noted in the ticket

### DIFF
--- a/sparkplug_plugin/sparkplug_b_datatype_test.go
+++ b/sparkplug_plugin/sparkplug_b_datatype_test.go
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// ENG-5126: Sparkplug B metric datatype must survive from BIRTH to DATA.
+// Sparkplug B metric datatype must survive from BIRTH to DATA.
 //
 // Per the Sparkplug B spec, NBIRTH/DBIRTH carries alias + name + datatype while
 // NDATA/DDATA carries only alias + value; the host is required to remember the
@@ -36,7 +36,7 @@ import (
 	"github.com/united-manufacturing-hub/benthos-umh/sparkplug_plugin/sparkplugb"
 )
 
-var _ = Describe("Datatype persistence across BIRTH and DATA (ENG-5126)", func() {
+var _ = Describe("Datatype persistence across BIRTH and DATA", func() {
 	Describe("AliasCache", func() {
 		var cache *sparkplugplugin.AliasCache
 
--- a/sparkplug_plugin/sparkplug_b_datatype_test.go
+++ b/sparkplug_plugin/sparkplug_b_datatype_test.go
@@ -37,7 +37,6 @@ import (
 )
 
 var _ = Describe("Datatype persistence across BIRTH and DATA (ENG-5126)", func() {
-
 	Describe("AliasCache", func() {
 		var cache *sparkplugplugin.AliasCache
 
@@ -191,7 +190,8 @@ var _ = Describe("Datatype persistence across BIRTH and DATA (ENG-5126)", func()
 			}
 		}
 
-		DescribeTable("decodes signed integer wire values using the BIRTH datatype",
+		DescribeTable(
+			"decodes signed integer wire values using the BIRTH datatype",
 			func(datatype uint32, wire interface{}, expected float64) {
 				birthMetric := &sparkplugb.Payload_Metric{
 					Name:     stringPtr("temp"),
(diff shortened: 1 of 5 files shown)


sparkplug_plugin/sparkplug_b_input.go
```diff
@@ -1614,6 +1611,36 @@ func (s *sparkplugInput) extractMetricValueRaw(metric *sparkplugb.Payload_Metric
 	return nil
 }
 
+// decodeIntValue reinterprets the unsigned int_value wire field according to the
+// metric datatype. Sparkplug B packs signed integers (Int8/Int16/Int32) as
+// two's-complement into int_value (ENG-5126); without a signed datatype the raw
+// uint32 is returned unchanged (UInt8/UInt16/UInt32, or datatype unknown).
+func decodeIntValue(datatype *uint32, raw uint32) interface{} {
```
led0nk: (nitty) can we in general check that we use `any` for interface calls here ?

sparkplug_plugin/sparkplug_b_datatype_test.go
```diff
@@ -0,0 +1,265 @@
...
+					Value: &sparkplugb.Payload_Metric_DoubleValue{DoubleValue: 1.5},
+				},
+			}
+			Expect(cache.ResolveAliases("Factory/Line1", dataMetrics)).To(Equal(1))
+			Expect(*dataMetrics[0].Name).To(Equal("legacy"))
+			Expect(dataMetrics[0].Datatype).To(BeNil())
+		})
+	})
+
+	Describe("NBIRTH→NDATA flow through the input", func() {
```
led0nk, line 115: generally speaking is this also something for DDATA? just wondering

led0nk (approved):
lgtm, just the nitpick and the question, if it's something you consider as `needs fix` then feel free to go for it, otherwise free to merge

sparkplug_plugin/sparkplug_b_datatype_test.go
Aaron99B: Yes, it's the same flow, but I added a spec for DDATA as well

sparkplug_plugin/sparkplug_b_input.go
Aaron99B: Changed that

PR #369, opened 2026-06-10
feat: preserve go struct order in schema export

FalconTube:
- Exported schemas now use the same order, as their golang struct
- Was verified by generating both schemas and just diffed them vs old ones


led0nk (approved):
lgtm with included gosnowflake downgrade

PR #374, opened 2026-06-16
fix: types for inline plugins now match (eng-4926)

led0nk:
# Description

Some plugins (e.g. `processor_bloblang`), which only use inline code are wrongly set to `tpye: object` in our json-schema.
This is used via CodeMirror to display linting-issues. 
But Plugins, which use these inline-code are in fact no `objects` but should be defined as `type: string`, which would be more correct. Therefore we guard plugins, which have no configuration fields and state them as no `objects` but instead as `string`.

This then resolves the linting issue, which is not supposed to happen in the first place.

grafik

---
### Important notice:

This will only be resolved after the next release of `benthos-umh`, which will then re-generate the schema.


JeremyTheocharis (changes requested):
Can you add a test showing e.g., the mentioned bloblang example that you are expecting a string there and not an object?

led0nk:
added test-cases as well as fixed 2 failing tests & Makefile command

JeremyTheocharis (approved):

PR #376, opened 2026-06-18
chore(deps): bump gopcua to v0.9.0

JeremyTheocharis:
## What this does

Bumps `github.com/gopcua/opcua` from `v0.8.0` to `v0.9.0`. Drop-in: no plugin code changes were needed (build + vet clean). The OPC UA input/output inherit a year of upstream client fixes.

Validated against live hardware before bumping (see the test-results comment below): on a real Kepware KEPServerEX, `v0.8.0` dropped an encrypted 1000-tag subscription with `EOF`, while `v0.9.0` subscribes cleanly. The full CI unit suite is green.

Changelog entries are added as changie fragments under `.changelog/unreleased/` (the `(client)` runtime items below are the ones our users get automatically).

## Full gopcua v0.9.0 changelog

The release ships a client and a server. Each item is tagged **(client)** or **(server)**; **API** items only matter to Go code written against the library. Through benthos-umh we use the client, so the plain **(client)** items are the runtime behavior we inherit.

### New

- `Basic256Sha256` channels now complete a handshake end to end in both `Sign` and `SignAndEncrypt` mode between a gopcua client and server, so you can run signed and encrypted connections (client + server)
- A `StateChangedFunc` callback lets a client run your own code on every connection-state change (Connected, Connecting, Disconnected, Reconnecting) (client, API)

### Improvements

- Monitoring a batch of nodes now succeeds for the valid ones instead of failing the whole batch when one node ID is bad; each rejected node is reported individually as a `monitor.ItemError` you can read with `errors.As` (client)
- Node operations now run against a client interface, so you can inject a mocked client when unit-testing code that calls node methods (client, API)

### Fixes

- Reading or subscribing to large sets of nodes (roughly a thousand or more) over an encrypted connection no longer drops the connection with an `EOF`; smaller requests and unencrypted connections were not affected (client + server)
- Subscriptions now survive a reconnect: after a network interruption the client no longer loses its subscriptions and stops receiving data until it is restarted (client)
- Recovering missed data after a reconnect no longer hangs in an endless retry loop or silently drops the recovered notifications (client)
- Connecting to servers that present their certificate as a chain of leaf plus intermediate certificates, such as Siemens WinCC Unified Runtime, now works (client)
- Browsing now returns a usable node ID for references typed as `ExpandedNodeID` instead of nil, which previously caused a panic (client)
- Calling `Close` on a client that never established a session no longer panics (client)
- expvar counters (read, send, dial, subscription) render again under `/debug/vars` on Go 1.23 (client, API)
- A server no longer crashes when a client writes to a node in a namespace that does not exist; it returns `Bad_NodeIdUnknown` (server)
- Server writes to non-Value attributes such as Description or DisplayName now report success instead of a false bad-attribute error (server)
- Server responses larger than the negotiated chunk size are now split across multiple chunks instead of dropping the connection (server)
- A server now returns a valid, non-zero session timeout, so strict clients such as UaExpert no longer fail to connect with `BadTimeout` (server)
- Creating a folder node in a server address space no longer panics (server)
- A server now sets the `ReceiverCertificateThumbprint` on its handshake response, so strict clients (Prosys, UaExpert) accept encrypted channels against a gopcua server (server)

**Full gopcua changelog**: https://github.com/gopcua/opcua/compare/v0.8.0...v0.9.0-rc1

### DIFF
--- a/.changelog/unreleased/fixes-20260618-163816.yaml
+++ b/.changelog/unreleased/fixes-20260618-163816.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: 'OPC UA: reading or subscribing to large tag sets (roughly a thousand or more) over an encrypted connection no longer drops the connection with an `EOF` error'
+time: 2026-06-18T16:38:16.000000+02:00
--- a/.changelog/unreleased/fixes-20260618-163817.yaml
+++ b/.changelog/unreleased/fixes-20260618-163817.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: 'OPC UA: subscriptions now survive a reconnect, so data collection resumes automatically after a network interruption instead of stalling until the input is restarted'
+time: 2026-06-18T16:38:17.000000+02:00
--- a/.changelog/unreleased/fixes-20260618-163818.yaml
+++ b/.changelog/unreleased/fixes-20260618-163818.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: 'OPC UA: connecting to servers that present their certificate as a chain (leaf plus intermediates), such as Siemens WinCC Unified, now works'
+time: 2026-06-18T16:38:18.000000+02:00
--- a/.changelog/unreleased/fixes-20260618-163819.yaml
+++ b/.changelog/unreleased/fixes-20260618-163819.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: 'OPC UA: a single unreadable or invalid node ID no longer blocks monitoring the rest of a subscribed batch'
+time: 2026-06-18T16:38:19.000000+02:00
--- a/.changelog/unreleased/fixes-20260618-163820.yaml
+++ b/.changelog/unreleased/fixes-20260618-163820.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: 'OPC UA: browsing certain servers no longer crashes the input'
+time: 2026-06-18T16:38:20.000000+02:00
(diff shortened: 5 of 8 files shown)


JeremyTheocharis:
## Hardware validation of gopcua v0.9.0

Before bumping, I validated the release against real OPC UA servers (not simulators), focused on the headline upstream fix: large encrypted subscriptions no longer dropping with `EOF` (gopcua [#824](https://github.com/gopcua/opcua/pull/824) / [#781](https://github.com/gopcua/opcua/issues/781)).

### Reproduced and fixed: Kepware KEPServerEX

Identical config on both versions — encrypted `SignAndEncrypt` / `Basic256Sha256`, subscribing to 1000 tags in one batch:

| Step | gopcua v0.8.0 | gopcua v0.9.0 |
|------|---------------|---------------|
| Encrypted connect | connected | connected |
| Monitor 1000 nodes (single `CreateMonitoredItems`) | **`Failed to monitor batch 0-999: EOF`** | **`Successfully monitored 1000 nodes`** |

This is the exact bug customers hit on encrypted connections with large tag sets: the 1000-item request response overflows the 64 KB secure-channel chunk, and v0.8.0 mishandled the chunking. v0.9.0 fixes it. Clean before/after on real hardware.

### Regression check: Ignition Gateway (Eclipse Milo)

Connected encrypted, browsed ~44k nodes, and monitored cleanly on both versions at our Ignition batch size (100). No regression. No `EOF` on either.

### Side finding (not part of this bump): Ignition batch size

Our `ignition` server profile caps `MaxBatchSize` at 100 while the code comments suggest 1000. I tested raising it: on v0.9.0, a 1000-item `CreateMonitoredItems` against Ignition returns `BadTimeout` — the server (Milo, 64 concurrent ops/session) can't create 1000 monitored items in one call fast enough, whereas Kepware did the same 1000 in ~2s. So the Ignition cap is partly a server-latency limit, independent of the gopcua fix; the real sweet spot is somewhere above 100 but below 1000. Tracking separately as a profile-tuning follow-up — out of scope for this dependency bump.

### CI

Full unit suite green (`test-opcua` and all other plugin unit jobs). PLC integration tests run in the merge queue.

led0nk (approved):

PR #377, opened 2026-06-19
test(sparkplug): pin NBIRTH/DBIRTH → UMH metadata contract (ENG-5176)

Aaron99B:
## Problem

[ENG-5176](https://linear.app/united-manufacturing-hub/issue/ENG-5176/managementconsole-default-sparkplug-b-template-drops-all): in Sparkplug B, a device sends a metric's **initial value** in its BIRTH certificate (NBIRTH/DBIRTH) and only sends **later changes** as DATA (NDATA/DDATA). UMH already processes DATA, but the default ManagementConsole template ignored BIRTH — so any metric that never changes after birth (most static and config tags) never reached the UNS, and customers saw an incomplete set of tags in the Topic Browser. A customer hit exactly this: Ignition showed the full tag tree, UMH showed only the actively-changing subset.

The ManagementConsole fix relaxes the template filter to let BIRTH through. That only works if the input already attaches full UMH metadata to birth metrics — which is the contract this PR guards.

## What

Unit specs that push a node-level NBIRTH and a device-level DBIRTH payload through the real `ProcessBirthMessage` + `CreateSplitMessages` path and assert each produced message carries:

- `umh_conversion_status == "success"`
- `umh_location_path` (populated)
- `umh_tag_name` (and `umh_virtual_path` when the metric name is hierarchical)
- the parsed payload value (`12.5`, `"v1.2"`)

## Why it was missing

Existing tests didn't pin this: `sparkplug_b_datatype_test.go` feeds a BIRTH but asserts only the value, and the `umh_*` integration test filters BIRTH out and validates DDATA only. Without this guard, a future input change could silently drop `umh_*` from birth metrics and re-break the customer scenario.

## Notes

- Test-only — no behavior change, no changelog fragment.
- `umh_location_path` is asserted non-empty rather than pinned, so the related edge-node placement fix ([ENG-5175](https://linear.app/united-manufacturing-hub/issue/ENG-5175/sparkplug-b-input-edge-node-dropped-from-location-path-devices-land-at)) can refine the mapping without breaking this test.

### DIFF
--- a/sparkplug_plugin/sparkplug_b_birth_umh_metadata_test.go
+++ b/sparkplug_plugin/sparkplug_b_birth_umh_metadata_test.go
@@ -1,0 +1,133 @@
+//go:build !integration
+
+// Copyright 2026 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NBIRTH/DBIRTH metrics must parse into full UMH metadata, not just a payload value.
+//
+// The ManagementConsole default Sparkplug B template (ENG-5176) relies on birth
+// metrics carrying umh_location_path / umh_tag_name / umh_virtual_path: it lets BIRTH
+// through so every metric declared in a birth certificate seeds the Topic Browser
+// with its initial value, instead of only the metrics that actively change via DATA.
+//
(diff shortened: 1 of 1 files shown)


sparkplug_plugin/sparkplug_b_birth_umh_metadata_test.go
JeremyTheocharis, line 1: Could we make the problem statement more explicit here and in the PR description? The key regression is customer-facing: in Sparkplug B, devices send initial metric values in BIRTH messages and later changes in DATA messages. We already process DATA, but if BIRTH is ignored, any metric that never changes afterwards never reaches the UNS, so customers see an incomplete set of tags. Framing the test around that failure mode would make it much clearer why this contract matters.

Aaron99B: Done in fc8c24c. The test doc comment now leads with the customer-facing failure mode you described — a device sends a metric's initial value in BIRTH and later changes in DATA, so ignoring BIRTH leaves any never-changing metric missing from the UNS and customers see an incomplete tag set. Also reframed the PR description around the same problem statement.

JeremyTheocharis (approved):

PR #388, opened 2026-06-23
feat: persistant cache

led0nk:
# Description

This PR introduces the noderedjs-cache to a persistant state. We use `bbolt` here, which already brings a lot in regard of a key/value store, which gets saved in a database-file.

In regards of concurrent access we have 2 patterns here:

- at first we guard this via the `cache.update` function, which can always be used to check on the existing value from a certain key and make sure it exists or has a certain value
- on top of that we can / will provide thread-safetiness via `umh-core`, where we can specify threads in the processing part to `threads: 1`, which will then be an additional guard on top

The combination of both makes sure that we don't have multiple messages writing to the same cache `key`. Other than that `bbolt` itself restricts multiple access to the database as well, meaning there will always be only 1 tx at a time.


nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -564,22 +626,64 @@ if (msg.payload.value <= 100 && alarmed) {
...
+			Examples("memory", "persistent"),
+		service.NewStringField("path").
+			Description("File path for the 'persistent' backend. Relative paths resolve against the working directory. Leading '~' expands to the home directory.").
+			Default("./cache.db"),
+		service.NewDurationField("ttl").
+			Description("Time-to-live for cached entries. After this duration without an overwrite, the entry is removed. Set to 0 to disable expiration.").
+			Default("1h"),
+	).
+		Description("Cache configuration for state across messages.").
+		Advanced())
```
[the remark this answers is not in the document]
led0nk: made this optional

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -564,22 +626,64 @@ if (msg.payload.value <= 100 && alarmed) {
...
+
 	wrappedCode := fmt.Sprintf(`
 		(function(){
 			'use strict';
 			%s
 		})()
 	`, code)
 
-	return NewNodeREDJSProcessor(wrappedCode, mgr.Logger(), mgr.Metrics(), cache.NewMemoryStore(0))
+	return NewNodeREDJSProcessor(wrappedCode, mgr.Logger(), mgr.Metrics(), store)
```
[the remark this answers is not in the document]
led0nk: addressd

nodered_js_plugin/cache/bbolt.go
```diff
@@ -0,0 +1,310 @@
...
+			if err == nil && !item.Expired() {
+				old = item.Value
+				exists = true
+			}
+		}
+
+		newVal, err := fn(old, exists)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
led0nk: documented as relevant for multi-access

nodered_js_plugin/cache/memory.go
```diff
@@ -100,6 +100,48 @@ func (m *MemoryStore) Delete(_ context.Context, key string) error {
...
+	item, ok := m.items[key]
+	if ok && !item.Expired() {
+		old = item.Value
+		exists = true
+	}
+
+	newVal, err := fn(old, exists)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
led0nk: doc

nodered_js_plugin/cache/cache.go
```diff
@@ -24,6 +30,12 @@ type Cache interface {
 	Get(ctx context.Context, key string) (any, bool)
 	// Delete removes the entry for key. No-op when key does not exist.
 	Delete(ctx context.Context, key string) error
+	// Update atomically reads, transforms, and writes a value under key.
```
JeremyTheocharis: Could we add a short explanation of when to use `cache.update` vs `cache.set`? `cache.set` is for simply storing a value, while `cache.update` is for cases where the new value depends on what is already stored. For example, a counter should use `cache.update`, because the next value depends on the current one. That would make the purpose of `update` much easier to understand.

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -564,22 +626,70 @@ if (msg.payload.value <= 100 && alarmed) {
...
+		service.NewStringField("backend").
+			Description("Cache backend. 'memory' is in-process and lost on restart. 'persistent' writes to a file on disk and survives restarts.").
+			Default("memory").
+			Examples("memory", "persistent"),
+		service.NewStringField("path").
+			Description("File path for the 'persistent' backend. Relative paths resolve against the working directory. Leading '~' expands to the home directory.").
+			Default("./cache.db"),
+		service.NewDurationField("ttl").
+			Description("Time-to-live for cached entries. After this duration without an overwrite, the entry is removed. Set to 0 to disable expiration.").
+			Default("1h"),
```
JeremyTheocharis: shouldnt the default be 0?

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -564,22 +626,70 @@ if (msg.payload.value <= 100 && alarmed) {
...
 }
-return msg;`))
+return msg;`)).
+	Field(service.NewObjectField("cache",
+		service.NewStringField("backend").
+			Description("Cache backend. 'memory' is in-process and lost on restart. 'persistent' writes to a file on disk and survives restarts.").
+			Default("memory").
+			Examples("memory", "persistent"),
+		service.NewStringField("path").
+			Description("File path for the 'persistent' backend. Relative paths resolve against the working directory. Leading '~' expands to the home directory.").
```
JeremyTheocharis: what does working directory mean here for benthos? the directory of the benthos binary?

docs/processing/javascript-api.md
```diff
@@ -134,8 +218,24 @@ cache.set("last_event_ms", Date.now());
 return msg;
 ```
 
+### Concurrency
```
JeremyTheocharis: i think there are two types of parallelism here:

1. the same processor running in parallel (solved with pipeline.threads: 1)
2. multiple processors accessing the same cache

The last part would still be possible right? how is a messgae processed? one message through the entire pipeline first, and then the next one? or is it per batch?

could it ever happen that message 1 goes through processor 1, then message 2 is processed in processor 1 and then message 1 in processed in processor 2. this might be unexpected for the user and has influence on the cache design (e.g., if someone is using the cache to exchange information between processors)

JeremyTheocharis (changes requested):
product review (didnt review the code yet)

nodered_js_plugin/cache/cache.go
JeremyTheocharis: also it woudl be helpful to describe the flow of data usign cache.set and get for a counter, and where there would be a problem. and then also how to use the function fn there. took me some time to get it.

Yes — `fn func(old any, exists bool)` is basically “the code that decides the new value”.

When `cache.update("count", fn)` runs, it does this:

* looks up the current value for `"count"`
* passes that value into `fn`
* stores whatever `fn` returns as the new value

So:

* `old` = the current stored value
* `exists` = whether there was already a value there

That lets you write logic like “if there is already a count, add 1, otherwise start at 1”.

With `set`/`get`, a counter would look like this:

```js
var count = 0;
if (cache.exists("count")) {
  count = cache.get("count");
}
count = count + 1;
cache.set("count", count);
```

This looks fine, but the problem is that these are separate steps.

If two messages run at the same time, this can happen:

* message A reads `count = 5`
* message B also reads `count = 5`
* message A adds 1 and writes `6`
* message B adds 1 and writes `6`

Now one increment is lost. You expected `7`, but got `6`.

With `update`, the counter looks like this:

```js
cache.update("count", function(old, exists) {
  if (exists) {
    return old + 1;
  }
  return 1;
});
```

Here the read and write belong together. The store says: “I’ll read the current value, let you calculate the next one, and write it back before someone else can change it.”

So if two messages increment at the same time:

* first one sees `5`, writes `6`
* second one then sees `6`, writes `7`

No increment is lost.

So the purpose of `fn(old, exists)` is:

* give you the current value safely
* let you decide the next value
* avoid mistakes when multiple messages update the same key at once

If you want, I can turn that into a clean second review comment you can post.

JeremyTheocharis: also similar hwat you have in the user facing docs, but just shorter with one example and then referencing the documentation. then i find it becomes more intuitive. a

led0nk: Expanded the interface doc on `Cache.Update` to call out when to use it (counters/toggles — new value depends on old) versus `Set` (independent overwrite), and pointed at `docs/processing/javascript-api.md` for the full counter-flow walkthrough you sketched. Kept the Go comment terse (project style) and put the longer narrative — race example with `get`+`set` losing increments, fixed via `cache.update` — into the user-facing docs so the explanation lives where users actually look.

nodered_js_plugin/nodered_js_plugin.go
led0nk: Yep — flipped the TTL default to `0s` (no expiration). Users who actually want auto-expiry can opt in with e.g. `ttl: 1h`. `0` matches the mental model better, especially for the `persistent` backend where surprise-deletion is worse than unbounded growth.

nodered_js_plugin/nodered_js_plugin.go
led0nk: Good catch — the term was ambiguous. Reworded the field description: relative paths resolve against the directory where the benthos process was started, and under UMH Core that's the S6 service directory for that DFC instance. Recommended absolute paths in the description to avoid the question entirely.

docs/processing/javascript-api.md
led0nk: Great question — both forms of parallelism exist; here's how each is handled now.

**1. Same processor in parallel (`pipeline.threads > 1`)**: N goroutines call `ProcessBatch` on the same processor instance concurrently. The cache is shared across those goroutines and serialized by the store mutex; `cache.update` is the atomic primitive that prevents the lost-update race. `threads: 1` also works but throttles throughput.

**2. Multiple processors sharing one cache**: this is now supported via a `cache.name` field plus an in-process refcounted registry. Two `nodered_js` processors with the same `backend` + `name` resolve to the same underlying store instance (via `cache.Acquire`), so a key set by proc1 is visible to proc2. Default `name` is `shared`, so the common case — two processors, no explicit cache config — already shares state out of the box. Set a different name to isolate groups, or empty (`name: ""`) to opt back into per-processor caches. For `persistent`, only the first processor needs `path`; later ones attaching by name omit it. Documented in the new "Sharing across processors" section of `docs/processing/javascript-api.md`.

**Message ordering**: with `threads > 1` there's no global ordering guarantee — proc2 can run on msg2 before proc1 finishes msg1, etc. But that doesn't break cache safety: every cache op (including the read+modify+write inside `cache.update`) is serialized by the store mutex, regardless of which thread or processor issued it.

**Out of scope**: two separate benthos PIDs hitting the same bbolt file — bbolt's flock blocks that, intentionally. For cross-process state, an external KV is the right tool.

JeremyTheocharis: what about message ordering with threads=1 how we have it in umh-core? i am thinking what is the best way to recommend to a user and whether we should nudge the user in using the update function over set/get more often? edepends on how the behaviour is with threads=1. if it is ordered in this case and there is only one processor, then the user doesn't have to use cache update right? if one of the other cases unordered or multiple processors, then he should use not set anymore and instead go for update (or otherwise model it)

JeremyTheocharis (changes requested):
as discussed in person (doing it so it gets out of my review queue)

JeremyTheocharis (changes requested):
See also: [Proposal: making cache.set / cache.get safe (threads=1, bridges vs. standalones, future API)](https://linear.app/united-manufacturing-hub/document/proposal-making-cacheset-cacheget-safe-threads1-bridges-vs-standalones-f6a415b63457)

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+
+At-least-once inputs (Kafka, UNS input, MQTT QoS1+, AMQP, JetStream) redeliver the same message when a downstream ACK is lost. Without protection, every retry re-runs the JavaScript, so a counter kept in the cache double-counts, an alarm re-fires, and a monotonic max advances past values it has already recorded.
+
+Both processors accept a `dedupKey` config field that names a **Benthos message metadata field** (set by the input plugin) whose value uniquely identifies the source event across retries. On the first sight of a value, cache writes run normally and the value is recorded in the cache under the reserved prefix `__dedup__:`. On later deliveries of the same value, `cache.set` and `cache.delete` are no-ops for that message, so cache state stays idempotent under retries.
+
+The dedup value is pulled from message metadata via `msg.meta.<dedupKey>`. It must be attached by the input plugin (for example the Kafka input auto-tags every message with `kafka_offset`, `kafka_partition`, `kafka_topic`). If your input doesn't attach a unique identifier, compose one with an upstream `bloblang` or `mapping` processor before the JS processor runs.
```
JeremyTheocharis, line 131: what happens if we want to use the timestamp_ms as dedup key?

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+# tag_processor
+tag_processor:
+  dedupKey: kafka_offset
+```
+
+**What it covers**
+
+- `cache.set` — no-op on retry
+- `cache.delete` — no-op on retry
+- Read-modify-write patterns become idempotent: the JS still runs and computes locally, but the retried write does not commit
```
JeremyTheocharis, line 148: but then what can happen is on a retry:<br><br>assume oyu ghave the code

x= cache.get("counter")

x++

cache.set("counter", x)

msg.payload.value = x

you would not update the cache, but the second produced message would still have an invalid counter with a double increased x.

Potential solution:

* in this cache, load the full message from the cache that was computed at that point in time?
* ?

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+- Multiple cache operations in one message run as one uninterrupted block
+- Cross-processor read-modify-write on a shared cache is also atomic
+
+Plain `get` / `set` is safe for counters, buffers, alarms, and all stateful patterns below.
+
+Trade-off: cache operations serialize on a shared mutex. For workloads under 100 msg/s, the cost is negligible.
+
+The auto-lock does not guarantee message order. With `pipeline.threads > 1`, messages process out of order; message 5 may arrive before message 3. For strict message order, set `pipeline.threads: 1`.
+
+### Idempotency and monotonicity under retries (`dedupKey`)
```
JeremyTheocharis, line 125: shouldnt we also need to validate that the incomign data on that dedup key is monotonic? would also help with the downsampler. otherwise it should error and throw that message away.

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+- `cache.delete` — no-op on retry
+- Read-modify-write patterns become idempotent: the JS still runs and computes locally, but the retried write does not commit
+- Monotonic counters, high-water marks, alarm latches — all stay correct across retries when built on `cache.set`
+
+Every suppressed message emits a WARN log line (`"cache: suppressing writes for retried message (dedupKey=... value=...)"`) and increments the `cache_dedup_suppressed` counter. Retries should be rare, so a WARN is intentional; a sustained high rate on this counter or a flood of these log lines usually means the upstream retry source is misbehaving.
+
+**What it does NOT cover**
+
+- Side effects outside the cache — HTTP calls, external DB writes, log lines, and metrics `Incr` still run on every retry
+- Output-side deduplication — messages sent to Kafka/UNS/MQTT still leave the pipeline on every retry. If your output needs dedup, gate it separately (e.g. Kafka idempotent producer, `umh_topic`-keyed compaction, or a downstream dedup filter)
```
JeremyTheocharis, line 156: coudl also be renamed "it is still at-least-once", so messages can be duplciated and leave duplicated the pipeline

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+- Messages missing the `dedupKey` meta field — a one-time warning is logged and the message is processed as fresh
```
JeremyTheocharis, line 157: should error hard instead and throw the message away

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+- Cross-process retries — dedup markers live in the same cache instance as your user keys. Two benthos processes must share the same persistent cache (`backend: persistent`, same `name`) to share dedup state, and bbolt's single-writer file lock still applies
+
+**How to use**
+
+1. Pick a metadata field whose value uniquely identifies the source event across retries. For Kafka this is the `topic:partition:offset` triple — `kafka_offset` alone repeats across partitions and topics.
```
JeremyTheocharis, line 162: Pick a metadata field or field in the payload that is constantly increasing. We will validate that messages come in there not out of order and will drop them otherwise

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+2. If the input doesn't already attach such a field (or attaches only its parts), compose one earlier in the pipeline with a `mapping`/`bloblang` step and write it to a fresh meta key.
+3. Set `dedupKey` to that meta key on the processor.
+4. Bound the marker lifetime with `cache.ttl`. Markers live under `__dedup__:<value>` in the same cache; a TTL matching your worst-case retry window (minutes for Kafka, hours for slow-consumer scenarios) keeps memory bounded. Without a TTL, markers accumulate for the lifetime of the process (or the persistent file).
```
JeremyTheocharis, line 165: no need to have a TTL for a marker? it is more optional and also doesnt take a lot of memory?

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+  kafka:
+    addresses: ["broker:9092"]
+    topics: ["events"]
+pipeline:
+  processors:
+    - mapping: |
+        meta dedup_id = meta("kafka_topic") + ":" + meta("kafka_partition") + ":" + meta("kafka_offset")
+    - nodered_js:
+        cache:
+          dedupKey: dedup_id
```
JeremyTheocharis, line 204: how about allowing this mapping straight in here for the dedupKey? there is an easy way to add this to a benthos plugin, a lot of other plugins do it as well

docs/processing/javascript-api.md
```diff
@@ -60,6 +108,124 @@ if (cache.exists("counter")) {
...
+        code: |
+          var n = cache.exists("count") ? cache.get("count") : 0;
+          cache.set("count", n + 1);
+          msg.payload = { count: n + 1 };
+          return msg;
+```
+
+- First delivery of offset 42: `count` was 41. JS reads 41 and writes 42. Payload reports `42`.
+- Redelivery of offset 42 (crash before ACK): JS reads 42 and computes 43, but the `cache.set("count", 43)` call is suppressed. The cache stays at 42. The retried message still leaves the pipeline with `43` in its payload; downstream systems must dedup output separately. Nothing kept in the cache is corrupted.
```
JeremyTheocharis, line 213: lets address this

JeremyTheocharis (changes requested):
see also direct message

docs/processing/javascript-api.md
JeremyTheocharis: because of this, we should probably gate this behind a FF or something?

JeremyTheocharis: or maybe behind the "cache: persistent" fag or something

docs/processing/javascript-api.md
JeremyTheocharis: for example in a stream processor

JeremyTheocharis:
pls note that the direct message and the draft PR is more up to date than the comments here.

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -493,16 +518,41 @@ func RecordDrop(counter *service.MetricCounter, logger *service.Logger, reason s
...
+	}
+	u.cache.Unlock()
+}
+
 func (u *NodeREDJSProcessor) ProcessBatch(ctx context.Context, batch service.MessageBatch) ([]service.MessageBatch, error) {
+	err := u.cacheBegin(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer u.cacheCommit(ctx)
```
JeremyTheocharis: this means we always commit even if errors appear?

docs/processing/javascript-api.md
```diff
@@ -39,106 +39,220 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
...
 ```javascript
-var prev = null;
-if (cache.exists("last_value")) {
-  prev = cache.get("last_value");
-}
 var delta = 0;
-if (prev !== null) {
-  delta = msg.payload.value - prev;
+if (cache.exists("last")) {
+  delta = msg.payload.value - cache.get("last");
```
JeremyTheocharis: will this not now cause issues upon replays?

so assume this message is replayed, then only the cache.set is blocked. but delta is calculated based on the newer value. so the emitted payload might be incorrect?

docs/processing/javascript-api.md
```diff
@@ -39,106 +39,220 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
...
+```
+
+### API
 
 ```javascript
-cache.set(key, value)    // Store a value under key (string)
-cache.get(key)           // Retrieve a value, logs error if key not found
-cache.exists(key)        // Returns true if key exists, false otherwise
-cache.delete(key)        // Remove a key
+cache.set(key, msg)   // Write msg.value only when msg.timestamp_ms is strictly newer than the stored one.
```
JeremyTheocharis: this is now very intuitive for time-series, but what about relational data? for relational data we could fallback to kafka_offset.

Alternative (to stay close with your choosen approach):

* cache.set(key, value, watermark) where watermark (wording tbd) can be either 
  * msg.payload.timestamp_ms or 
  * msg.meta.kafka_offset

Also we should explain the case where two incoming umh_topics write into the same key. This does not work here. Therefore, a cache can only be updated from one incoming umh_topic (one continious monotonic stream).

docs/processing/javascript-api.md
```diff
@@ -39,106 +39,220 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
...
+```
+cache.set: dropped stale write for key "k" (timestamp_ms=... not newer than stored)
+```
+
+Retries should be rare in a healthy pipeline. A sustained flood of these WARN lines means the upstream input is replaying more than expected — investigate there.
+
+**What it covers**
+
+- `cache.set` under retries and out-of-order arrivals — the store is the source of truth.
+- Counters, high-water marks, alarm latches, history buffers built on `cache.set` — all stay consistent.
```
JeremyTheocharis: what is a high-water mark?

docs/processing/javascript-api.md
```diff
@@ -39,106 +39,220 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
...
+
+- `cache.get` → compute → `cache.set` in the same message is atomic: no other message can slip in between the get and the set
+- Multiple cache operations in one message run as one uninterrupted block
+- Cross-processor read-modify-write on a shared cache is also atomic
+
+Trade-off: cache operations serialize on a shared mutex. For workloads under 100 msg/s, the cost is negligible.
+
+The auto-lock does not guarantee message order. With `pipeline.threads > 1`, messages process out of order; message 5 may arrive before message 3. The timestamp gate on `cache.set` is what keeps state correct under reordering, not the mutex. For strict message order set `pipeline.threads: 1`.
+
+### Examples (time-series payloads)
```
JeremyTheocharis, line 182: maybe we add for these examples tests that show the out of order behaviour? Not onyl checking in the cache, but also the actual output of those in nodered_js?

See also the tests that I added in my PR.

docs/processing/javascript-api.md
```diff
@@ -39,106 +39,220 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
...
+Always use `cache.exists(key)` before `cache.get(key)` to avoid error logs on missing keys:
+
+```javascript
+if (cache.exists("last_temperature")) {
+  var last = cache.get("last_temperature");
 }
 ```
 
-### Counter
+### Timestamp gating: replays and out-of-order writes
```
JeremyTheocharis, line 100: What was about ACID? To explain how we ensure (or not ensure) them?

* Atomic
* Consistency
* Isolation
* Durability

I remember we had there some thoughts in our previous meetings.

Especially how we ensure atomic updates across kafka and the cache here.

docs/processing/javascript-api.md
```diff
@@ -39,106 +39,220 @@ Accepts multiple arguments: `console.log("value is", msg.payload.value)`
...
+
+`cache.set` is the same idempotency primitive under two names:
+
+- **Replay**: same source event redelivered by the input (Kafka after ACK loss, MQTT QoS1+ retry, UNS input restart). The redelivered message carries the same `timestamp_ms` it did the first time, so the second `set` is a no-op.
+- **Out-of-order**: a newer message arrived first, then an older one shows up. The older `timestamp_ms` is not strictly greater than the stored one, so the write is dropped.
+
+Both cases produce one WARN log line:
+
+```
+cache.set: dropped stale write for key "k" (timestamp_ms=... not newer than stored)
```
JeremyTheocharis: can we make these error messages more user facing? e.g., 'dropped write for key "k" (current watermark "1234" is older than the last processed "2222". Incoming messages have to be in order, so either messages are being replayed or there is an issue with the watermark coming from the data source (e.g., time went backwards)'

JeremyTheocharis (changes requested):

docs/processing/javascript-api.md
JeremyTheocharis: something different than timestamp_ms is neededf though as otherwise

timestamp_ms=msg.meta.kafka_offset will be really confusing

docs/processing/javascript-api.md
led0nk: rephrased that

docs/processing/javascript-api.md
led0nk: put that to watermark which accepts both, timestamp_ms as well as e.g. kafka_offset

nodered_js_plugin/nodered_js_plugin.go
led0nk: yes, we do - per-message js errors will drop the poisoned message before and if anything happens on the transaction itself it will get handled via bbolt. Other successful messages in the batch shouldn't be rolled back here

i renamed this to `cacheEnd` which is more precise here

docs/processing/javascript-api.md
led0nk: did that and packed this to errors.go

docs/processing/javascript-api.md
led0nk: added as we discussed a part for explanation how we handle acid

docs/processing/javascript-api.md
led0nk: added tests for those (also adjusted the examples a bit due to refinements)

docs/processing/javascript-api.md
led0nk: this is fixed now, it required a change as you now are getting back not only the value with `cache.get` but as well the `watermark` you previously provided, with this you can make it fail-proven

docs/processing/javascript-api.md
led0nk: also added something regarding the only one umh_topic issue

PR #390, opened 2026-06-24
sparkplug_b input: add passthrough_raw_metric flag (ENG-5242)

Aaron99B:
## Background

The `sparkplug_b` input decodes the standard Sparkplug B payload but drops anything carried in proto2 extension fields. The plugin's generated types don't declare those fields, so they never reach the decoded value. The bytes aren't lost on the wire (proto2 retains unknown fields); they're just never exposed.

Some Sparkplug B publishers carry data in those extensions and need it decoded inside UMH-core rather than in an external service.

## What this adds

An opt-in `passthrough_raw_metric` flag on the input. When enabled, each per-metric message also carries the raw, re-marshaled `Payload.Metric` as base64 metadata `spb_metric_raw`. A downstream protobuf decoder that has the device's `.proto` (ENG-5243's `nodered_js` `protobuf.decode`, or the Redpanda `protobuf` processor) reads the extension fields from there.

- The **whole** metric is marshaled, not the top-level `GetUnknown()` fragment: an extension may be nested inside a sub-message such as `MetaData`, and the decoder needs a complete `Payload.Metric`.
- Off by default; output is byte-for-byte unchanged when disabled.
- Best-effort: a marshal error attaches nothing rather than dropping the message.

Deliverable 1 of 2 for ENG-5229; the downstream decoder is ENG-5243.

## Tests

`passthrough_raw_metric_test.go`: the enabled case round-trips the metric and asserts a nested `MetaData` extension survives marshal→unmarshal via `spb_metric_raw`; the disabled case asserts standard metadata and payload are untouched and no key is added.

## For reviewers

The raw bytes travel to Kafka headers downstream and roughly double per-metric size, so very large metrics may approach broker message/header size limits. Called out in the field description and `docs/input/sparkplug-b-input.md`. Opt-in, off by default.

### DIFF
--- a/.changelog/unreleased/new-20260624-151831.yaml
+++ b/.changelog/unreleased/new-20260624-151831.yaml
@@ -1,0 +1,3 @@
+kind: new
+body: Sparkplug B input can attach the raw proto-encoded metric as base64 metadata (`spb_metric_raw`) via the `passthrough_raw_metric` flag, exposing proto2 extension fields for downstream decoding (ENG-5242)
+time: 2026-06-24T15:18:31.361763+02:00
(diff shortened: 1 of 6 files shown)


JeremyTheocharis (commented):
lgtm from product

PR #391, opened 2026-06-24
nodered_js: add protobuf.decode/encode with inline descriptor set (ENG-5298)

Aaron99B:
## What this adds

A `protobuf` namespace in the JS runtime, so bridges that ingest protobuf can decode and re-encode messages inside a JS processing step — reaching fields the standard inputs don't decode. The schema is passed inline as a base64 `FileDescriptorSet` (compile once with `protoc --include_imports --descriptor_set_out`), because no proto compiler ships in the binary and the schema can't be a file on disk (no file-injection; airgapped installs have no toolchain).

```javascript
protobuf.decode(dataB64, descriptorSetB64, msgName)  // -> object
protobuf.encode(obj, descriptorSetB64, msgName)      // -> base64 string
```

- Extension types are registered, so proto2 extensions surface as protojson `[pkg.ext]` keys.
- Both throw a catchable JS error on failure.
- Available in both `nodered_js` and `tag_processor` (they share `SetupJSEnvironment`).
- Built registries are memoized by descriptor set.

Tracked by ENG-5298.

## Background

This started as deliverable 2 of the Sparkplug B extension work (ENG-5229), which needed a JS-side way to decode proto2 extension fields. That decode moved inside the `sparkplug_b` input directly (PR #392), so a JS protobuf surface is no longer needed for Sparkplug B. It stands on its own, though: any bridge ingesting protobuf can use it. Shipping it as a general `nodered_js` feature, decoupled from Sparkplug B.

## Security

The descriptor set and data are untrusted input from a user script, so:

- The registry cache is bounded (128 entries, eviction) — an unbounded cache would let a script exhaust process memory by decoding with many distinct descriptor sets.
- The JS-facing `decode`/`encode` closures recover any Go panic into a returned error — goja does not recover Go panics, so without this a malformed descriptor hitting a latent panic would crash the whole process.

## Tests

`nodered_js_plugin/protobuf/`: round-trip with top-level and message-nested extensions; error paths (bad base64 / unknown or non-message name / malformed descriptor / non-object encode); memo transparency; the cache bound. A goja-level test proves decode works through the JS boundary and that errors throw. gremlins mutation testing reports 100% efficacy on the package.

## For reviewers

No proto compiler ships — the user supplies a precompiled descriptor set. protojson conventions apply to the decoded object (int64/bytes as strings, enums as names), documented in `docs/processing/javascript-api.md`.

### DIFF
--- a/.changelog/unreleased/new-20260624-160831.yaml
+++ b/.changelog/unreleased/new-20260624-160831.yaml
@@ -1,0 +1,3 @@
+kind: new
+body: Node-RED JS and tag processor expose a `protobuf` namespace (`protobuf.decode` / `protobuf.encode`) to decode and encode protobuf messages inline using an embedded base64 descriptor set, including proto2 extension fields (ENG-5243)
+time: 2026-06-24T16:08:31.625342+02:00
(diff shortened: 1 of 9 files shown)


JeremyTheocharis (commented):
OK from product, awaiting @led0nk for code review

JeremyTheocharis (approved):

PR #392, opened 2026-06-25
Sparkplug B: decode proto2 extension fields from an inline schema (ENG-5229)

Aaron99B:
## What

Opt-in decoding of proto2 Sparkplug extension fields in the `sparkplug_b` input. Some devices carry data in proto2 extension fields on `Payload.MetaData` (per-metric metadata) and `Payload.MetricValueExtension` (the metric value). The standard decode retains those bytes but can't name them, because they aren't in the compiled schema. This exposes them, decoded in-platform, with no second flow. See ENG-5229 for the originating use case.

## How it works

`decode_extensions.extensions` takes an inline proto2 snippet — `package` + `extend` blocks (plus any custom message types or well-known-type imports):

```yaml
input:
  sparkplug_b:
    mqtt: { urls: ["tcp://broker:1883"] }
    identity: { group_id: "factory_a" }
    role: "primary"
    decode_extensions:
      extensions: |
        package acme;
        extend org.eclipse.tahu.protobuf.Payload.MetaData {
          optional int64 extra_value = 9;
        }
```

At startup the plugin compiles the snippet once (`bufbuild/protocompile`), linked against the embedded Sparkplug descriptor so the `extend` resolves to the exact message the input decodes — measured at ~0.5 ms, then reused read-only. Per metric it re-marshals (proto2 retains the extension bytes), decodes against the generated `Payload.Metric` with the extension resolver, and attaches:

- `spb_ext_<field>` — each scalar extension (e.g. `spb_ext_extra_value`), readable with no protobuf knowledge downstream;
- `spb_metric_decoded` — the full decoded metric as JSON, the catch-all for message-typed extensions, lifted with one `JSON.parse` in the tag_processor.

Only metrics that carry an extension get these keys; everything else is byte-for-byte unchanged.

The snippet must be proto2 (`extend` is a proto2-only feature; the plugin injects the `syntax`/`import` and rejects a customer-supplied one). Startup fails with a clear message on an unparseable snippet, a snippet with no extensions, or two scalar extensions whose names would collide into the same `spb_ext_*` key.

## Consuming it in the tag_processor

The decoded fields are plain metadata, read in the bridge's `tag_processor`. For the `extra_value` schema above:

```yaml
pipeline:
  processors:
    - tag_processor:
        defaults: |
          msg.meta.location_path = "factory_a.line1." + msg.meta.spb_device_id;
          msg.meta.data_contract = "_historian";
          msg.meta.tag_name = msg.meta.spb_metric_name;

          // Scalar extension: read the flat key directly (int64 arrives as a string).
          if (msg.meta.spb_ext_extra_value) {
            msg.payload.extra_value = Number(msg.meta.spb_ext_extra_value);
          }

          // Message-typed extension: parse the JSON blob; keys are the bracketed FQN,
          // e.g. dec.extensionValue["[acme.my_message]"] or dec.metadata["[acme.extra_value]"].
          if (msg.meta.spb_metric_decoded) {
            const dec = JSON.parse(msg.meta.spb_metric_decoded);
            // lift whatever fields you need from dec ...
          }
          return msg;
```

Both keys are present only on metrics that carry an extension, so the guards also skip ordinary metrics.

## Notes for reviewers

- **Additive contract.** The decode is appended to the end of `createMessageFromMetric` and only calls `MetaSet`. It does not touch `extractMetricValue`, the existing `spb_*` metadata, the alias cache, birth/death, or sequence tracking. With the feature off, behaviour is unchanged.
- **`MetricValueExtension` is nested under `Payload`, not `Payload.Metric`** — the extendee FQN is `org.eclipse.tahu.protobuf.Payload.MetricValueExtension`.
- `bufbuild/protocompile` was already a transitive dependency; this promotes it to a direct require.

## Testing

- `make test-sparkplug`: 292 passed. 8 new specs cover the scalar round-trip, the message-typed-to-JSON path, the no-extension case, and the four startup validations.
- The decode path (proto2 retention through the generated struct → re-marshal → resolver decode) was validated end-to-end against device-shaped wire bytes before implementation.
- Not yet run: the local-bridge e2e (a live MQTT message into the UNS).

## Out of scope / follow-ups

- ManagementConsole ergonomics (a snippet editor / template) could come later; the benthos side is complete.

### DIFF
--- a/sparkplug_plugin/extension_decode.go
+++ b/sparkplug_plugin/extension_decode.go
@@ -38,11 +38,14 @@ const (
 	// resolver cannot satisfy the injected import.
 	extSparkplugImportPath = "proto/sparkplug_b_official.proto"
 	// extPreamble is prepended so the snippet compiles as proto2 and the Sparkplug extendees
-	// resolve. extPreambleLines must track the number of newlines so diagnostics map back to
-	// the customer's line.
-	extPreamble      = `syntax = "proto2"; import "` + extSparkplugImportPath + `";` + "\n"
-	extPreambleLines = 1
+	// resolve.
+	extPreamble = `syntax = "proto2"; import "` + extSparkplugImportPath + `";` + "\n"
 )
+
+// extPreambleLines is the number of lines extPreamble adds ahead of the customer's snippet, so
+// remapCompileError can subtract them and map diagnostics back to the customer's line. Derived
+// from extPreamble so it can't drift if the preamble changes.
+var extPreambleLines = strings.Count(extPreamble, "\n")
 
 // extensionDecoder decodes the proto2 extension bytes a Sparkplug metric carries but the
 // standard decode ignores. It is built once and is read-only afterwards.
@@ -91,10 +94,31 @@ func newExtensionDecoder(snippet string) (*extensionDecoder, error) {
 	}, nil
 }
 
+// carriesExtension reports whether the standard decode left unrecognized bytes on a message
+// decode resolves extensions against. proto2 retains extension fields as unknown bytes, so an
+// empty result means there is definitely nothing to decode. The check is a necessary, not
+// sufficient, condition: any registered extension always lands here, but unrelated unknown
+// fields can too, so decode still confirms before emitting. It must cover the same messages as
+// supportedExtendees.
+func carriesExtension(metric *sparkplugb.Payload_Metric) bool {
+	if md := metric.GetMetadata(); md != nil && len(md.ProtoReflect().GetUnknown()) > 0 {
+		return true
+	}
+	if ve := metric.GetExtensionValue(); ve != nil && len(ve.ProtoReflect().GetUnknown()) > 0 {
+		return true
+	}
+	return false
+}
+
 // decode returns the metric's scalar extensions as leaf->value pairs and the full decoded
 // metric as JSON; present is false when the metric carries no extension. Re-marshaling
 // recovers the extension bytes that proto2 retained through the standard decode.
 func (d *extensionDecoder) decode(metric *sparkplugb.Payload_Metric) (map[string]string, string, bool, error) {
+	// Skip the re-marshal/decode for the common case of a metric with no extension bytes.
+	if !carriesExtension(metric) {
+		return nil, "", false, nil
+	}
+
 	raw, err := proto.Marshal(metric)
 	if err != nil {
 		return nil, "", false, fmt.Errorf("re-marshal metric: %w", err)
@@ -131,6 +155,7 @@ func collectExtensions(m protoreflect.Message, flat map[string]string) bool {
 		if fd.Kind() == protoreflect.MessageKind || fd.Kind() == protoreflect.GroupKind {
 			switch {
 			case fd.IsMap():
+				// Maps don't carry extensions in the Sparkplug schema, so there's nothing to recurse into.
 			case fd.IsList():
 				l := v.List()
 				for i := 0; i < l.Len(); i++ {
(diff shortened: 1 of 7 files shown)


JeremyTheocharis (commented):
lgtm from product

sparkplug_plugin/extension_decode.go
```diff
@@ -0,0 +1,295 @@
...
+	return &extensionDecoder{
+		types:      types,
+		metricDesc: (&sparkplugb.Payload_Metric{}).ProtoReflect().Descriptor(),
+	}, nil
+}
+
+// decode returns the metric's scalar extensions as leaf->value pairs and the full decoded
+// metric as JSON; present is false when the metric carries no extension. Re-marshaling
+// recovers the extension bytes that proto2 retained through the standard decode.
+func (d *extensionDecoder) decode(metric *sparkplugb.Payload_Metric) (map[string]string, string, bool, error) {
```
led0nk, line 116: can we guard this so it doesn't run on each metric, but instead already check on e.g. the extension carry?

sparkplug_plugin/extension_decode.go
```diff
@@ -0,0 +1,295 @@
...
+const (
+	extSnippetVirtualPath = "extensions.proto"
+	// extSparkplugImportPath must equal the path recorded in the embedded descriptor, or the
+	// resolver cannot satisfy the injected import.
+	extSparkplugImportPath = "proto/sparkplug_b_official.proto"
+	// extPreamble is prepended so the snippet compiles as proto2 and the Sparkplug extendees
+	// resolve. extPreambleLines must track the number of newlines so diagnostics map back to
+	// the customer's line.
+	extPreamble      = `syntax = "proto2"; import "` + extSparkplugImportPath + `";` + "\n"
+	extPreambleLines = 1
```
led0nk: does it make sense to just count this in a variable?

sparkplug_plugin/extension_decode.go
```diff
@@ -0,0 +1,295 @@
...
+		if ev := fd.Enum().Values().ByNumber(v.Enum()); ev != nil {
+			return string(ev.Name())
+		}
+		return strconv.FormatInt(int64(v.Enum()), 10)
+	default:
+		return v.String()
+	}
+}
+
+func sanitizeExtLeaf(name string) string {
```
led0nk, line 220: somehow looks duplicated like `sanitizeForTopic` can you check?

sparkplug_plugin/extension_decode.go
```diff
@@ -0,0 +1,295 @@
...
+	m.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
+		if fd.IsExtension() {
+			present = true
+			if !fd.IsList() && !fd.IsMap() && isScalarExtKind(fd.Kind()) {
+				flat[sanitizeExtLeaf(string(fd.Name()))] = scalarToString(fd, v)
+			}
+		}
+		if fd.Kind() == protoreflect.MessageKind || fd.Kind() == protoreflect.GroupKind {
+			switch {
+			case fd.IsMap():
```
led0nk, line 157: lets have a minor comment here or do some handling whenever we simply do nothing here

led0nk (changes requested):

sparkplug_plugin/extension_decode.go
Aaron99B: Done

sparkplug_plugin/extension_decode.go
Aaron99B: Yes, changed that now

sparkplug_plugin/extension_decode.go
Aaron99B: Good catch. I moved both to a sanitiseRunes function

sparkplug_plugin/extension_decode.go
Aaron99B: Included that

led0nk (approved):

PR #393, opened 2026-06-25
uns_beta_single: single-plugin candidate alongside uns_beta (option B, ENG-5105)

JeremyTheocharis:
### What this is

A second, additive UNS consumer input, `uns_beta_single`, registered alongside the existing `uns_beta` (the template form from #367). Both coexist; nothing in #367 is modified or deleted. The production `uns_beta` name stays with the template form until the comparison is decided.

`uns_beta_single` does in one Go registration what `uns_beta` does in two (a template + a reader): it translates the lean 4-field user surface into a full redpanda config, builds connect's `redpanda` input itself in Go, and applies the umh_topics key filter, ack-gating, and metadata aliasing. The lean surface (`consumer_group` + `umh_topics` required; `broker_address` default `localhost:9092`; `kafka_topic` default `umh.messages`) is identical to `uns_beta` by design, so the two forms are comparable on behavior, not on config.

Note on reuse: `uns_beta_single` does not wrap the existing `newUnsBetaInputFor`/`unsBetaInput`. It reimplements the key filter, the all-filtered self-ack, and the `umh_topic`/`kafka_msg_key` aliasing inline in its own `ReadBatch` loop, because the forwarder returns a `service.BatchInput` (no `ConnectionTest`) rather than the `readBatcher` the wrapper expects. This inline reimplementation is what the parameterized suite checks against `uns_beta`.

### Why two forms exist, and what this tests

`uns_beta` is two registrations because of a Go `internal/` wall: connect's redpanda constructor (`NewFranzReaderToggledFromConfig`) and its field helpers live in `connect/v4/internal/impl/kafka`, unimportable from benthos-umh. The template form routes around it by letting benthos build the redpanda input as a managed child from a parsed config field. This PR tests the alternative: lift the wall with a **forwarder shim** (a small package patched in under `connect/v4/` that can import `internal/impl/kafka`), so a single Go input can construct the redpanda input directly.

The open question this PR exists to answer: is the single-plugin form worth it? It removes the two-registration parity-by-convention, but it pays for that with a compile-time-mandatory patch and coupling to connect's internal `init()` assembly. The decision is deferred until the two forms are compared; this PR does not replace #367.

### How it works

- **Forwarder shim** (`patches/connect-redpanda-forward.patch`): creates `connect/v4/public/components/kafka/redpandaforward/`, exporting `ConfigFields()` (delegates to a new `RedpandaInputConfigFields()` wrapper added to `internal/impl/kafka`, so a structural change in connect propagates through the shim, so no manual mirror can drift) and `NewRedpandaInput(conf, res)` (re-assembles connect's redpanda input `init()`: connection details + consumer opts + the reader + the three wrappers). Explicit-connection-only; the shared-redpanda-client fallback is omitted (unused by this caller, since no benthos-umh caller registers `SharedGlobalRedpandaClientKey`).
- **`uns_beta_single` Go input**: 4 lean fields → redpanda config YAML pinned to the same 12 values `uns_beta`'s template pins → `ParseYAML` → `NewRedpandaInput(conf, res)` with the outer stream's resources → inline filter/ack/alias loop (see reuse note above).
- **`make patch-connect`** now applies both patches (key-filter + forwarder) sequentially, each with its own `git apply --check` drift gate.

### What's proven

- **`redpanda_lag` reaches the outer stream's metrics registry** under Go-side construction (verified at runtime, not just at the source level). The gauge binds to `res.Metrics()` at `franz_reader_ordered.go:534`, regardless of construction path, so the single-plugin form keeps the metrics the two-plugin rationale feared losing.
- **Behavioral parity (parameterized suite)**: 11 behavioral scenarios run against both forms via a plugin-name parameter: delivery, metadata contract, key filter, all-filtered commit, start-from-earliest, mixed-batch NACK redelivery, gapless restart, unreachable broker, NACK capstone, plus the `dropped_keyless` and `filtered_records` counter scenarios. All green for both forms. A behavior fix to either form is checked against the other; drift between them is caught.
- **Translation parity**: all 12 pinned redpanda values reproduced exactly, asserted against the Go fetch constants (not magic strings) and cross-checked against `newBetaKeyFilter`, plus a field-for-field comparison against `uns_beta`'s rendered template.
- **Perf parity**: the E2E drain benchmark runs `select_one` (200k records, 1 kept) against `uns_beta`, `uns_beta_single`, and legacy `uns`. Both `uns_beta` forms land at ~412k allocs/op; the `key_pattern` source pre-filter omits 199,999 of 200,000 records before message construction in both. (The kfake broker can produce noisy high-alloc outliers for `uns_beta` from its eager `FieldInput` kgo client thrashing against the fake broker; the benchmark header warns kfake `uns_beta` numbers are not throughput-trustworthy. The real-broker `BenchmarkE2E_DrainRedpanda` is the throughput comparison of record.)

### How this differs from #367

| | `uns_beta` (#367, template) | `uns_beta_single` (this PR, option B) |
|---|---|---|
| Registrations | Two (template alias + Go reader) | One (Go input) |
| Config translation | Template (YAML, at config-load) | Go (at construction) |
| Inner input construction | Eager (framework `FieldInput` builds the redpanda input at config-load) | Lazy (built in `Connect`) |
| Connect coupling | Public config schema only | Internal `init()` assembly (via shim) |
| Patch | key-filter, perf-only / runtime-optional | key-filter + forwarder, compile-mandatory |
| Parity enforcement | By convention (render tests) | By Go construction + parameterized suite |
| `redpanda_lag` | Reaches outer stream (managed child) | Reaches outer stream (outer `res` passed in); confirmed at runtime |

Both forms deliver identical records, metadata, and redelivery guarantees across the 11 parameterized scenarios. Perf parity is measured, not assumed: the E2E drain benchmark runs `select_one` (200k records, 1 kept) against both, and both land at ~412k allocs/op. The `key_pattern` source pre-filter (the build-time `connect-key-filter.patch`, inherited from #367) omits non-matching records before `service.Message` construction identically in both forms; the plugin-level filter is defense-in-depth. The one known behavioral divergence is the spoofed-keyless counter (below).

### Known gap (documented, deferred)

The parameterized suite found one real divergence: `uns_beta`'s `classifyDrop` is spoof-first (a record with a foreign `kafka_key` header shadowing an empty native key is tallied as `dropped_spoofed_key`), while `uns_beta_single` lacks `hasSpoofHeader` and mis-counts such records as `dropped_keyless`. `uns_beta_single` also does not register `dropped_spoofed_key`. Spoof detection is a tracked, deferred workstream (**ENG-5125**); this PR documents the gap with a `TODO(ENG-5125)` at the divergence point rather than porting the logic separately. ENG-5125 will port `hasSpoofHeader`/`classifyDrop` to the single form (and rework it for both). Narrow exposure (mixed-producer topics with an external `kafka_key` header injector). The `dropped_spoofed_key` counter scenario is left `uns_beta`-only for that reason; the other 11 scenarios are parameterized.

### What this is NOT

- Not replacing #367. #367 stays the likely-ship; this is a comparison candidate.
- Not a clean (patch-free) form. The clean single-plugin form needs an upstream connect re-export of `NewFranzReaderToggledFromConfig` + the field helpers from `public/components/kafka`, filed as a follow-up, not as part of this PR. That re-export retires the **forwarder patch**. The **key-filter patch** retires separately, only if connect upstream adopts a `key_pattern` feature (a different upstream contribution); the two patches retire under different conditions.
- Not user-visible / no changelog. `uns_beta_single` is an internal comparison candidate behind a patched-only build; the changelog waits for the commit that picks between the two forms.
- Not porting spoof detection; deferred to ENG-5125.

### Reviewer attention

The forwarder patch is compile-mandatory for this branch (the unpatched lane doesn't have `uns_beta_single`). Local dev requires `make patch-connect` before `go test`/`go build`. The patched CI lane (`test-uns-patched`, kfake, no docker) is the green signal. The branch is 4 commits behind staging and will need a rebase before merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -241,3 +241,5 @@ cmd/schema-export/schema-export
 .connect-patched/
 go.patched.mod
 go.patched.sum
+003_klaus/
+vsdd/
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ CONFIG ?= ./config/opcua-hex-test.yaml
 # copy is git-baselined so the patch can always be regenerated with `git diff`.
 CONNECT_PKG := github.com/redpanda-data/connect/v4
 CONNECT_PATCH := patches/connect-key-filter.patch
+CONNECT_FORWARD_PATCH := patches/connect-redpanda-forward.patch
 CONNECT_VENDOR := .connect-patched
 PATCHED_MODFILE := go.patched.mod
 PATCHED_SUMFILE := go.patched.sum
@@ -78,14 +79,18 @@ patch-connect:
 	( cd $(CONNECT_VENDOR) && git init -q && git add -A && \
 	  git -c user.email=patch@umh -c user.name=patch commit -qm base && \
 	  git apply --check -p1 ../$(CONNECT_PATCH) && \
-	  git apply -p1 ../$(CONNECT_PATCH) ) && \
+	  git apply -p1 ../$(CONNECT_PATCH) && \
+	  git apply --check -p1 ../$(CONNECT_FORWARD_PATCH) && \
+	  git apply -p1 ../$(CONNECT_FORWARD_PATCH) ) && \
 	cp go.mod $(PATCHED_MODFILE) && cp go.sum $(PATCHED_SUMFILE) && \
 	go mod edit -replace $(CONNECT_PKG)=./$(CONNECT_VENDOR) $(PATCHED_MODFILE) && \
 	echo "patched $(CONNECT_PKG) -> ./$(CONNECT_VENDOR); committed go.mod untouched." && \
 	echo "build/test patched with: GOFLAGS=-modfile=$(PATCHED_MODFILE) go test -tags connect_patched ./uns_plugin/..."
 
-# Drift gate only: verify the diff still applies to the pinned module without
-# touching go.mod. CI runs this so a Renovate bump that moves connect fails loud.
+# Drift gate only: verify the key-filter patch still applies to the pinned
+# module without touching go.mod. Not run by CI; run manually after a
+# connect bump. The forwarder patch is exercised by patch-connect (and the
+# connect_patched test lane), not by this gate.
 .PHONY: check-connect-patch
 check-connect-patch:
 	@CONNECT_SRC=$$(go list -m -f '{{.Dir}}' $(CONNECT_PKG)) && \
--- a/patches/connect-redpanda-forward.patch
+++ b/patches/connect-redpanda-forward.patch
@@ -1,11 +1,12 @@
 diff --git a/internal/impl/kafka/input_redpanda.go b/internal/impl/kafka/input_redpanda.go
+index b2956c6..0f772fb 100644
 --- a/internal/impl/kafka/input_redpanda.go
 +++ b/internal/impl/kafka/input_redpanda.go
-@@ -98,6 +98,13 @@
- 		},
+@@ -99,6 +99,14 @@ func redpandaInputConfigFields() []*service.ConfigField {
  	)
  }
  
++
 +// RedpandaInputConfigFields re-exports the redpanda input's config fields so
 +// packages outside internal/impl/kafka (the redpandaforward shim) can build a
 +// redpanda ConfigSpec without duplicating redpandaInputConfigFields.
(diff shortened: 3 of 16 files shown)


JeremyTheocharis:
Superseded by #395 — the converged single-form `uns_beta`. Aaron and Janik both preferred the single-plugin form (Option B) on 2026-06-26; the template form (#367) is dropped and the parameterized comparison suite is collapsed to single-form. The preference thread from this PR is captured in #395's body.

PR #398, opened 2026-06-28
feat(nodered_js): array fan-out + tag_processor metrics wiring (ENG-5240)

JeremyTheocharis:
## Background

`nodered_js` and `tag_processor` share the same goja JavaScript runtime, but only `tag_processor` accepted an array return to fan out one output per element. A read bridge that pulls a JSON array from an ERP API and needs one UNS message per record could not be built with `nodered_js` alone: a function returning an array errored the batch. The workaround (`nodered_js` then `unarchive` then `nodered_js`) is not expressible in the new UI and times out on deploy.

## Bug

Two gaps beyond the missing fan-out turned up while fixing it. First, `nodered_js` swallowed JavaScript execution errors per message: the message vanished from the output, was acknowledged, and never surfaced (silent data loss). `tag_processor` had the same class of swallow on condition errors. Second, both plugins serialized non-scalar metadata values into Kafka headers with `fmt.Sprintf("%v")`, so a nested null appeared as the literal string `<nil>` and a map appeared as Go-syntax text that downstream header consumers cannot parse.

## Fix

- `nodered_js` accepts an array return and fans out one output per element, mirroring `tag_processor`. Nil elements are skipped, non-object elements error, and an empty or all-null array drops the input.
- A throwing `nodered_js` function now forwards the errored message and continues the rest of the batch, so one bad message no longer discards the good ones produced before it; the error is logged and counted in `messages_errored`. (A JS throw, a non-object return, a bad array element, and an unrecoverable infrastructural failure like byte conversion all take this path.) This replaces a silent per-message swallow. Operators should alert on `messages_errored` for nodered_js failures.
- A throwing `tag_processor` condition now forwards the errored message to downstream error handling instead of swallowing it. Defaults and advanced errors remain batch-fatal.
- Both plugins share a new `SetMetaFromJS` helper that JSON-marshals non-scalar metadata (nested null becomes `null`, not `<nil>`) and stringifies scalars, with a `%v` fallback when a value cannot be marshaled (NaN or Infinity nested in a map).
- Counter ownership is corrected: `messages_dropped` is bumped once per drop, `messages_errored` once per errored message (not by batch size), and `messages_processed` counts successfully produced outputs. A drop and a throw in the same batch are independent per-message outcomes.

## Not shipping

- Visual multi-processor chain builder in the new UI (separate).
- Bridge deploy timeout fix (ENG-5238, separate).
- `tag_processor` defaults/advanced errors remain batch-fatal (not forwarded like conditions and nodered_js); aligning them is a follow-up.
- `SetMetaFromJS` falls back to Go-syntax `%v` for NaN/Infinity nested in a map/slice (non-empty, matches `tag_processor`'s `autoConvertValue`); a JSON-valid placeholder is a follow-up.
- 10k-record load test and mutation testing (follow-up PR).

Refs: ENG-5255 (parent), ENG-5240, ENG-5256 (canceled), ENG-5238 (out of scope).


nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -362,37 +362,94 @@ func (u *NodeREDJSProcessor) setupCache(ctx context.Context, vm *goja.Runtime) e
...
+			if err != nil {
+				// json.Marshal errors on NaN/+Inf nested in maps or slices;
+				// fall back to a non-empty value rather than writing an empty
+				// Kafka header (matches tag_processor's autoConvertValue).
+				newMsg.MetaSet(k, fmt.Sprintf("%v", val))
+				continue
 			}
+			newMsg.MetaSet(k, string(b))
+		default:
+			newMsg.MetaSet(k, fmt.Sprintf("%v", val))
```
led0nk, line 464: think it makes sense to use sth like `strconv.FormatFloat` here as numbers arrive in float64 and there might be issues with scientific notation here

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -362,37 +362,94 @@ func (u *NodeREDJSProcessor) setupCache(ctx context.Context, vm *goja.Runtime) e
...
+// JSON-marshaled so nested nil leaves become valid JSON null instead of
+// Go-syntax "<nil>". All other values are stringified with fmt %v. It is
+// exported so tag_processor can reuse this logic.
+func SetMetaFromJS(newMsg *service.Message, meta map[string]interface{}) {
+	for k, val := range meta {
+		if val == nil {
+			continue
+		}
+		switch val.(type) {
+		case map[string]interface{}, []interface{}:
```
led0nk: just something personal, could you run `go fix -any` for this PR's directories?

tag_processor_plugin/tag_processor_plugin.go
```diff
@@ -337,6 +352,15 @@ func (p *TagProcessor) ProcessBatch(ctx context.Context, batch service.MessageBa
 			continue
 		}
 
+		// An errored message (e.g. from a condition) is forwarded as-is,
+		// skipping validation and construction so it reaches the consumer
+		// marked errored and unchanged (its original payload preserved for
+		// downstream error handling/DLQ).
+		if msg.GetError() != nil {
```
led0nk: messages which have an error set are then not being constructed with `umh_topic`, is this meant to be like this, because this will likely cause issues for redpanda later on ?

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -432,17 +515,21 @@ func (u *NodeREDJSProcessor) ProcessBatch(ctx context.Context, batch service.Mes
 	return []service.MessageBatch{resultBatch}, nil
 }
 
-// processSingleMessage processes a single message using a VM from the pool
-func (u *NodeREDJSProcessor) processSingleMessage(ctx context.Context, msg *service.Message) (*service.Message, bool, error) {
+// processSingleMessage processes a single message using a VM from the pool.
```
led0nk: can we keep the comments short again? Not saying not neccessary, but lets focus on the comments to be like 1-2 lines maximum and absolutely on point. makes the whole thing more readable & focusable

tag_processor_plugin/tag_processor_plugin.go
```diff
@@ -309,9 +310,22 @@ func (p *TagProcessor) ProcessBatch(ctx context.Context, batch service.MessageBa
 		var newBatch service.MessageBatch
 
 		for _, msg := range batch {
+			// A message already errored by an earlier condition skips the
+			// remaining conditions and is forwarded unchanged.
```
led0nk: we should guard here for `if msg == nil` as well

nodered_js_plugin/nodered_js_plugin.go
```diff
@@ -362,37 +362,94 @@ func (u *NodeREDJSProcessor) setupCache(ctx context.Context, vm *goja.Runtime) e
...
+// messageFromReturnValue builds a service.Message from a single JS return
+// value, which must be a message object (a map with payload/meta).
+// NewMessage(nil) is safe: the engine's v2BatchedToV1Processor wrapper
+// restores the input context onto outputs in production, so
+// Copy()/WithContext() is a no-op.
+func messageFromReturnValue(v interface{}) (*service.Message, error) {
+	returnedMsg, ok := v.(map[string]interface{})
 	if !ok {
-		return service.NewMessage(nil), false, fmt.Errorf("function must return a message object or null")
+		return service.NewMessage(nil), fmt.Errorf("function must return a message object or null")
```
led0nk: lets just return nil,fmt.Errorf here, no need to create a message for this

led0nk (changes requested):
Is there meant to be a template later on in ManagementConsole although its documented now?

JeremyTheocharis:
Thanks Janik. #1 was right; reworked the error handling. Please re-review.

**#1:** you were right. An errored message forwarded without `umh_topic` can't be written to `umh.messages` (it requires `umh_topic` as the Kafka key), and the input here is usually a polled source (s7/modbus/opcua) with no backpressure and no buffer, so a message that can't be written blocks the poll loop and no new data comes in. Forwarding errored messages (this PR) and the old batch-fatal retry both hit this. Fix: drop poisoned messages at the processor instead (omitted from the batch, counted under `messages_dropped{reason=}`, Warn-logged). `ProcessBatch` no longer returns an error. The comment on `RecordDrop` has the detail; the capstone test feeds `[good, poisoned, good]` and asserts 2 written + 1 dropped, and fails if a drop is reverted to `SetError`+append.

A real DLQ (sentinel error topic) is a follow-up under ENG-5255, not this PR.

**#2–#7:** see inline. (#3 — no MC template planned; the array-return behavior is documented in `docs/processing/node-red-javascript-processor.md`.)

tag_processor_plugin/tag_processor_plugin.go
JeremyTheocharis: See the summary above. The comment on `RecordDrop` has the full engine trace.

nodered_js_plugin/nodered_js_plugin.go
JeremyTheocharis: Done. Returns `nil, err`; both callers discard the message on error.

nodered_js_plugin/nodered_js_plugin.go
JeremyTheocharis: Leaving as-is. `%v` on a float64 already produces the same notation `stringify()` and the payload use; `'f'` wouldn't match, and it renders 1e308 as a 309-digit string. Headers are just bytes, and `ParseFloat` handles scientific notation, so it round-trips fine. No change.

tag_processor_plugin/tag_processor_plugin.go
JeremyTheocharis: Done. Added `if msg == nil { continue }` to the conditions, `processMessageBatchWithProgram`, and nodered_js `ProcessBatch` loops, to match the guards already elsewhere in the file. Nil can't actually reach these loops; this just makes it consistent.

nodered_js_plugin/nodered_js_plugin.go
JeremyTheocharis: Done. Trimmed the long comments to 1-2 lines (`HandleExecutionResult` was 13). Kept the `RecordDrop` rationale, tightened.

nodered_js_plugin/nodered_js_plugin.go
JeremyTheocharis: Done. Ran `go fix`: `interface{}` to `any`, `for i := 0; i < N; i++` to `for range N` (kept `i` where it's used).

tag_processor_plugin/tag_processor_plugin.go
```diff
@@ -293,25 +291,25 @@ func (p *TagProcessor) ProcessBatch(ctx context.Context, batch service.MessageBa
...
-	// Process defaults with compiled program (Phase 2 optimization)
+	// Process defaults with compiled program
 	if p.defaultsProgram != nil {
-		var err error
-		batch, err = p.processMessageBatchWithProgram(ctx, batch, p.defaultsProgram, "defaults")
-		if err != nil {
-			return nil, fmt.Errorf("error in defaults processing: %w", err)
-		}
+		batch, _ = p.processMessageBatchWithProgram(ctx, batch, p.defaultsProgram, "defaults")
 	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Intentional — `batchFatalErr` was removed as part of the drop-loudly rework (see the summary at the top + the comment on `RecordDrop`). The old batch-fatal retry nacks the whole batch and stalls the polling input; per-message drop replaces it.

.changelog/unreleased/fixes-20260627-120101.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: Drops and JavaScript errors in the tag processor and Node-RED JavaScript processor now appear in the messages_dropped and messages_errored metrics instead of going unreported (ENG-5240)
```
led0nk: think this needs an update with `messages_errored`

.changelog/unreleased/fixes-20260627-120102.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: A Node-RED JavaScript processor function that throws now forwards the errored message and continues the rest of the batch, so one bad message no longer discards the good ones; the error is logged and counted in messages_errored instead of being silently dropped (ENG-5240)
```
led0nk: could now be something like this:
"...now drops the message loudly — omitted from the batch, warn-logged, counted in messages_dropped{reason=js_throw} — and the rest of the batch continues."

.changelog/unreleased/fixes-20260627-120103.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: A tag processor condition that throws now forwards the errored message to downstream error handling instead of silently dropping it (ENG-5240)
```
led0nk: could then be something like this:
"A tag processor condition that throws now drops the message loudly (messages_dropped{reason=js_throw}, warn-logged) instead of dropping it silently."

But i think we could get rid of one of those honestly

.changelog/unreleased/fixes-20260627-120105.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: The Node-RED JavaScript processor messages_processed counter now counts successfully produced outputs, and a throwing message is counted as errored rather than processed, instead of the counter counting every input attempt and inflating on retries (ENG-5240)
```
led0nk: also not coutned as errored rather again "dropped (messages_dropped{reason=js_throw})"

led0nk (approved):
rest looks good to me if you update the changelog entries here

JeremyTheocharis:
Post-approval changes, FYI in case you want another look. All in response to your changelog feedback plus a self-review pass.

- **Changelog + doc:** the Node-RED JS processor doc's throw/non-object-return section still described the reverted forward-on-error path ("forwarded marked with the error", "bridge goes degraded") — fixed to drop-loudly. Added a note that `messages_processed` counts outputs (drop-rate ratios shift under fan-out). Changelog entries now name `messages_dropped{reason=js_throw}`.
- **Non-map `meta` now drops loudly:** returning `{payload, meta:"str"}` (meta not an object) was silently skipped — the message proceeded without metadata. Now a `bad_return` drop via `RecordDrop`. nodered_js (single + array paths) and tag_processor.
- **VM pool:** `processMessageBatchWithProgram` releases the goja VM via `defer` in a per-message helper (was five manual `putVM` calls). No behavior change.
- **`messages_errored` removed** from nodered_js + tag_processor (dead since the drop-loudly rework). FYI changelog entry; umh-core doesn't consume this metric (its benthos parser only reads input_/output_/processor_ families), so no umh-core impact. `classic_to_core` and `sparkplug_b_input` still bump it — follow-up on ENG-5255.
- **Tests:** fixed a data race on the `messages` slice in tag_processor stream-builder tests (`go test -race` clean); added a positive `messages_processed` assertion.

Green: `go build ./...`, `make test-noderedjs`, `make test-tag-processor`, `go test -race` on both. Re-requesting your review for the non-changelog changes.

led0nk (approved):

PR #354, opened 2026-06-02
test(ads): Beckhoff ADS hardware integration tests (5/5)

RuneRoven:
## Summary

Adds hardware integration tests and unit tests. Depends on PR 4/5. Tests are skipped automatically when hardware env vars are not set.

- `beckhoff_ads_test.go`: TC2 + TC3 hardware tests — interval read, notification, route registration, reconnect, cycleTime verification, deterministic formula checks
- `beckhoff_ads_config_test.go`: config-driven pipeline tests via Benthos stream builder
- `deterministic_helpers_test.go`: exported PLC value formula helpers shared across test files
- `deterministic_helpers_unit_test.go`: unit tests for `CreateSymbolList`, `validateIP`, `validateAMSNetID`, `sanitize`, formula helpers — no hardware required
- `testdata/benthos-tc3.yaml`: TC3 pipeline config (38 symbols)
- `testdata/benthos-tc2.yaml`: TC2 pipeline config

**Running hardware tests:**
```bash
TEST_ADS_TC3_TARGET_IP=192.168.x.x \
TEST_ADS_TC3_TARGET_AMS=x.x.x.x.1.1 \
TEST_ADS_TC3_RUNTIME_PORT=851 \
TEST_ADS_TC3_ROUTE_USER=Administrator \
TEST_ADS_TC3_ROUTE_PASS=1 \
make test-ads
```

## Part of series
- 1/5 — init, config, connect
- 2/5 — CI workflow
- 3/5 — docs
- 4/5 — ReadBatch implementation (**base of this PR**)
- **5/5 — tests** ← this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #350, opened 2026-06-02
feat(ads): Beckhoff ADS plugin — init, config, connect (1/5)

RuneRoven:
## Summary

Adds the Beckhoff ADS input plugin foundation: dependency, struct, config, Connect, and Close. ReadBatch is a stub — full implementation in PR 4/5.

- `go-ads/v2 v2.2.0` dependency (direct TCP, no TwinCAT router)
- Plugin registration in `cmd/benthos/bundle` and schema-export
- `test-ads` Makefile target
- `AdsCommInput` struct + `PlcSymbol` + full config spec
- Config fields grouped: target connection, local AMS, route registration, read mode, advanced, symbols
- IP and AMS NetID validation at parse time
- `CreateSymbolList`: positional + `key=value` timing overrides per symbol (e.g. `GVL.var:maxDelay=0s:cycleTime=10ms`)
- Route registration via `WithRoute` (go-ads probe-first, UDP port 48899)
- Initial sample wait in `Connect` (notification mode) guarantees complete first batch
- `ReadBatch` stub satisfying `service.BatchInput` interface

## Part of series
- **1/5 — init, config, connect** ← this PR
- 2/5 — CI workflow
- 3/5 — docs
- 4/5 — ReadBatch implementation
- 5/5 — tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #400, opened 2026-07-01
feat(ads): Beckhoff ADS plugin — connectivity (2/6)

RuneRoven:
Connect, route creation, reconnect, Close. Adds the go-ads→benthos slog bridge (`benthosLogHandler`): trace suppressed, severities mapped, verbosity governed solely by the pipeline log level (no global `SetDefaultLogger`). Port fix: host-IP auto-detect TCP dial follows `targetPort` (UDP discovery stays 48899).

---

### DIFF
--- a/beckhoff_ads_plugin/ads.go
+++ b/beckhoff_ads_plugin/ads.go
@@ -52,7 +52,7 @@ type AdsCommInput struct {
 	client           Client
 	Log              *service.Logger
 	Symbols          []PlcSymbol
-	symbolByName     map[string]*PlcSymbol // configured symbol name → *PlcSymbol, populated in NewAdsCommInput
+	symbolByName     map[string]*PlcSymbol // configured symbol name → *PlcSymbol, populated in initSymbolIndex (Connect)
 	NotificationChan chan *Update
 	TransmissionMode int
 
@@ -217,11 +217,6 @@ func NewAdsCommInput(conf *service.ParsedConfig, mgr *service.Resources) (servic
 		mgr.Logger().Warnf("%s", w)
 	}
 
-	symbolByName := make(map[string]*PlcSymbol, len(symbolList))
-	for i := range symbolList {
-		symbolByName[symbolList[i].Name] = &symbolList[i]
-	}
-
 	m := &AdsCommInput{
 		TargetIP:         targetIP,
 		TargetAMS:        targetAMS,
@@ -233,7 +228,6 @@ func NewAdsCommInput(conf *service.ParsedConfig, mgr *service.Resources) (servic
 		MaxDelay:         maxDelay,
 		CycleTime:        cycleTime,
 		Symbols:          symbolList,
-		symbolByName:     symbolByName,
 		Log:              mgr.Logger(),
 		IntervalTime:     intervalTime,
 		RequestTimeout:   requestTimeout,
@@ -259,11 +253,97 @@ func init() {
 	}
 }
 
-// Connect, ReadBatch and Close are stubs; the go-ads adapter (Task 2) fills in
-// connection, read and shutdown behaviour behind the Client seam.
+// sessionConfig builds the library-agnostic connection spec passed to the
+// go-ads adapter.
+func (a *AdsCommInput) sessionConfig() SessionConfig {
+	return SessionConfig{
+		TargetIP:       a.TargetIP,
+		TargetAMS:      a.TargetAMS,
+		HostIP:         a.HostIP,
+		HostAMS:        a.HostAMS,
+		TargetPort:     a.TargetPort,
+		RuntimePort:    a.RuntimePort,
+		HostPort:       a.HostPort,
+		Username:       a.Username,
+		Password:       a.Password,
+		RequestTimeout: a.RequestTimeout,
+	}
+}
 
-func (a *AdsCommInput) Connect(_ context.Context) error {
-	return service.ErrNotConnected
+// targetSummary lists the PLC-side connection parameters for logging.
+func (a *AdsCommInput) targetSummary() string {
+	return fmt.Sprintf(
+		"  target IP:    %s\n"+
+			"  target port:  %d\n"+
+			"  target AMS:   %s\n"+
+			"  runtime port: %d",
+		a.TargetIP, a.TargetPort, a.TargetAMS, a.RuntimePort)
+}
+
+// hostSummary lists our own AMS identity for logging, marking the values that
+// are derived at runtime rather than configured.
+func (a *AdsCommInput) hostSummary() string {
+	hostAMS := a.HostAMS
+	if hostAMS == "auto" {
+		hostAMS = "auto (derived on connect)"
+	}
+	hostPort := fmt.Sprintf("%d", a.HostPort)
+	if a.HostPort == 0 {
+		hostPort = "random"
+	}
+	hostIP := a.HostIP
+	if hostIP == "" {
+		hostIP = "auto-detected"
+	}
+	return fmt.Sprintf(
+		"  host AMS:     %s\n"+
+			"  host port:    %s\n"+
+			"  host IP:      %s",
+		hostAMS, hostPort, hostIP)
+}
+
+// Connect establishes the ADS session, resolves symbol metadata, and (depending
+// on config) loads the full symbol table and registers notifications.
+func (a *AdsCommInput) Connect(ctx context.Context) error {
+	if a.client != nil {
+		return nil
+	}
+
+	c, err := newGoADSClient(ctx, a.sessionConfig(), a.Log)
+	if err != nil {
+		return err
+	}
+	a.client = c
+
+	if err = a.finishConnect(ctx); err != nil {
+		a.client.Close()
+		a.client = nil
+		return err
+	}
+	return nil
+}
+
+// finishConnect drives connect → symbol resolution → (optional) symbol table
+// load → (optional) notification setup; split out so Connect can clean up a.client on failure.
+func (a *AdsCommInput) finishConnect(ctx context.Context) error {
+	a.Log.Infof("Connecting to PLC:\n%s\n%s", a.targetSummary(), a.hostSummary())
+	if err := a.client.Connect(ctx); err != nil {
+		a.Log.Errorf("Connecting to PLC failed:\n%s\n  error:        %v", a.targetSummary(), err)
+		return err
+	}
+	a.Log.Infof("Connecting to PLC succeeded:\n%s", a.targetSummary())
+	a.initSymbolIndex(ctx)
+	if a.LoadSymbols {
+		if err := a.loadSymbolTable(ctx); err != nil {
+			return err
+		}
+	}
+	if a.ReadType == "notification" {
+		if err := a.setupNotifications(ctx); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (a *AdsCommInput) ReadBatch(_ context.Context) (service.MessageBatch, service.AckFunc, error) {
@@ -271,5 +351,17 @@ func (a *AdsCommInput) ReadBatch(_ context.Context) (service.MessageBatch, servi
 }
 
 func (a *AdsCommInput) Close(_ context.Context) error {
+	if a.client == nil {
+		return nil
+	}
+	a.Log.Infof("Closing connection to PLC:\n%s", a.targetSummary())
+
+	err := a.client.Close()
+	a.client = nil
+	if err != nil {
+		a.Log.Errorf("Closing connection to PLC failed:\n%s\n  error:        %v", a.targetSummary(), err)
+		return err
+	}
+	a.Log.Infof("Closing connection to PLC succeeded:\n%s", a.targetSummary())
 	return nil
 }
--- a/beckhoff_ads_plugin/client_goads.go
+++ b/beckhoff_ads_plugin/client_goads.go
@@ -1,0 +1,210 @@
+// Copyright 2026 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package beckhoff_ads_plugin
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net"
+	"net/netip"
+	"strconv"
+	"time"
+
+	adsLib "github.com/RuneRoven/go-ads/v2"
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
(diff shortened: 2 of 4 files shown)


RuneRoven:
@coderabbitai review

beckhoff_ads_plugin/ads.go
```diff
@@ -268,10 +268,67 @@ func init() {
...
+	// Ensure done channel is initialized (may be nil if constructed directly in tests)
+	if g.done == nil {
+		g.done = make(chan struct{})
+	}
+	g.Log.Infof("Creating new connection")
+	opts, err := g.buildSessionOptions(ctx)
+	if err != nil {
+		return err
+	}
+	if err := g.openSession(ctx, opts); err != nil {
```
led0nk: just style but lets keep the approach with first assigning and then checking on the error here as well

beckhoff_ads_plugin/ads.go
```diff
@@ -268,10 +268,67 @@ func init() {
...
+		if g.Handler != nil {
+			_ = g.Handler.Close()
```
led0nk: lets make heare use of something like 
```golang
        defer func() {
                if err != nil {
                        g.closeHandler()
                }
        }()
```

where then `g.closeHandler()` has the guard included itself

beckhoff_ads_plugin/connect.go
```diff
@@ -0,0 +1,243 @@
...
+	return opts, nil
+}
+
+// resolveRouteHostIP returns the local IP for the ADS route: the configured
+// hostIP, else the source IP of a TCP dial to the ADS port (same interface).
+func (g *AdsCommInput) resolveRouteHostIP(ctx context.Context) (string, error) {
+	if g.HostIP != "" {
+		return g.HostIP, nil
+	}
+	dialer := net.Dialer{Timeout: 3 * time.Second}
```
led0nk: for all those things i'd like to do a const-block somewhere - could also lilve in constants.go and then write a small (1 -line) coment why its 3 seconds

also for e.g. targetPort a few lines below the initial-samples and so on

beckhoff_ads_plugin/logging.go
```diff
@@ -0,0 +1,69 @@
...
+	attrs  []slog.Attr
+}
+
+func (h *benthosLogHandler) Enabled(_ context.Context, level slog.Level) bool {
+	// Suppress trace (LevelTrace = -8); forward debug and above to benthos,
+	// which then applies its own configured level filter.
+	return level >= slog.LevelDebug
+}
+
+func (h *benthosLogHandler) Handle(_ context.Context, r slog.Record) error {
```
led0nk, line 37: as said, we need to make sure to not introduce another logger here and make use of the benthos-umh one. If a function e.g. returns an error wwe could make use of that and use the benthos-logger but don't inject / wrap or do sth else with the logging

led0nk (changes requested):
i wrote a smole gist for this in general which would make the library on the later stage easier swappable. 

https://gist.github.com/led0nk/7e679c9ae206a56b3e4a37fc49d9672a

The reamde is generated but take a look more on the code, which was an idea from my perspective. The relevant part here is that we reconstruct some of the struct and make use of the interface so we can later on easily swap out things here. Then we also can redruce the points of imports for the library here. 

Btw some things might get better naming e.g. `goADSClient` could also simply be Client or sth like this

beckhoff_ads_plugin/logging.go
RuneRoven: go-ads is slog-based and always logs: if you pass no logger, client.go:136 defaults to slog.Default(). So dropping the bridge doesn't remove logging — go-ads's connection/route/notification diagnostics then spew to slog.Default() → raw stderr, bypassing benthos's pipeline and level control.

bridge → benthos routes the go-ads slog to g.Log with the benthos log level.

This simply bridges all logging direct into benthos to get the same format and not the stderr output.

beckhoff_ads_plugin/ads.go
RuneRoven: - closeHandler() (read.go) closes ASYNC — go func(){ _ = c.Close() }(). That's deliberate for the read hot-path (ReadBatchPull/ReadBatchNotification):
  go-ads Close() can block on protocol teardown, and a synchronous close there would stall the Benthos read loop. Async is right there.
  - Connect's failure cleanup closes SYNC — g.client.Close() directly. That's right for a setup path: on a failed connect Benthos will retry Connect, and
  we want the dead session fully torn down before returning, so the retry starts clean.
  
  (Connect→initSymbolIndex→loadSymbolTable→setupNotifications) into one finishConnect, so there's a single error path and one explicit close — already
  DRY, nothing to dedupe with a defer.
  - newGoADSClient failure needs no cleanup (client not yet assigned → still nil) — handled by only cleaning up after g.client = c.

beckhoff_ads_plugin/ads.go
```diff
@@ -263,17 +257,82 @@ func init() {
...
 
-func (g *AdsCommInput) Connect(_ context.Context) error {
-	return service.ErrNotConnected
+// Connect establishes the ADS session, resolves symbol metadata, and (depending
+// on config) loads the full symbol table and registers notifications.
+func (g *AdsCommInput) Connect(ctx context.Context) error {
+	if g.client != nil {
+		return nil
+	}
+	if g.done == nil {
```
led0nk: check if we really need to do this here, since we created the channel ourseelves earlier on, otherwise we could then only do it here e.g.

beckhoff_ads_plugin/client_goads.go
```diff
@@ -0,0 +1,208 @@
...
+
+// goADSClient is the go-ads adapter — the ONLY file (besides client.go)
+// importing the fork. All WithRoute/NewAMSAddress/type-mapping logic lives here.
+type goADSClient struct {
+	session *adsLib.Session
+}
+
+// isLikelyContainerIP flags private ranges commonly used by container/pod
+// networks, where route registration would advertise an unreachable IP.
+func isLikelyContainerIP(ip string) bool {
```
led0nk: let's challenge if we really need to have this here in the first place.

what if one comes up with an ip network starting with `10.0.0.0` here? 
we have it in the office and it would be recognized as potential container ip, then i'm questioning to have this in the first place. It makes sense to simply check the same way as in the configuration PR on the IP. So whether this is parsable to a IPv4 in the first place, but then i'd think of that it makes sense to reuse the one of the configuration PR

beckhoff_ads_plugin/ads.go
```diff
@@ -263,17 +257,82 @@ func init() {
...
+		g.client = nil
+		return err
+	}
+	return nil
+}
+
+// finishConnect drives connect → symbol resolution → (optional) symbol table
+// load → (optional) notification setup; split out so Connect can clean up g.client on failure.
+func (g *AdsCommInput) finishConnect(ctx context.Context) error {
+	g.Log.Infof("Connecting to plc")
```
led0nk: maybe let's print out more information here, e.g.:
- targetIP + port
- hostAMS + port
could also be multiline if it get's too long

beckhoff_ads_plugin/ads.go
```diff
@@ -263,17 +257,82 @@ func init() {
...
+		}
+	}
+	return nil
 }
 
 func (g *AdsCommInput) ReadBatch(_ context.Context) (service.MessageBatch, service.AckFunc, error) {
 	return nil, nil, service.ErrNotConnected
 }
 
 func (g *AdsCommInput) Close(_ context.Context) error {
```
led0nk: maybe for closing the connection we should as well state some information, such as (example):
`Closing connection to (ip-address):(port)`

`Connection successfully closed`

This seems a bit unusual but in later stages it sometimes helps on debugging

beckhoff_ads_plugin/client_goads.go
```diff
@@ -0,0 +1,208 @@
...
+	// CGNAT range used by some Kubernetes CNIs: 100.64.0.0/10
+	if p[0] == 100 && p[1] >= 64 && p[1] <= 127 {
+		return true
+	}
+	return false
+}
+
+// resolveRouteHostIP returns the local IP for the ADS route: the configured
+// hostIP, else the source IP of a TCP dial to the ADS port (same interface).
+func resolveRouteHostIP(ctx context.Context, cfg SessionConfig, log *service.Logger) (string, error) {
```
led0nk, line 37: just in order to understand this correctly here:

- we do a tcp-port dial
- > if it fails we do a udp-port dial
- > other than that we report back the ip address and append this with `1.1` outside of this? 

Is this the automatic route creation then?

beckhoff_ads_plugin/connect.go
```diff
@@ -0,0 +1,122 @@
...
+
+// loadSymbolTable downloads the full symbol and datatype table from the PLC;
+// required for struct/array symbols.
+func (g *AdsCommInput) loadSymbolTable(ctx context.Context) error {
+	g.Log.Infof("Loading symbol and datatype table from PLC (loadSymbols=true)")
+	if err := g.client.LoadSymbols(ctx); err != nil {
+		g.Log.Errorf("LoadSymbols failed: %v", err)
+		return err
+	}
+	g.Log.Infof("Symbol table loaded")
```
led0nk: Seems nitty but lets do the logging here consistent, maybe sth like:

Loading symbol and datatype table from PLC...

-> error case: `Loading symbol and datatype failed:...`

successcase: `Loading symbol and datatype succeeded.`

beckhoff_ads_plugin/connect.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// AddNotifications returns nil errors even when all symbols fail to
+	// resolve (e.g. PLC not yet ready); detect that here so Benthos retries Connect().
+	registered := 0
+	for _, r := range results {
+		switch {
+		case r.Registered:
+			registered++
+		case r.Skipped:
+			g.Log.Errorf("Notification symbol %q skipped (check symbol name)", r.SymbolName)
```
led0nk: if i understand correctly we error out if registered notifications are 0

then we will try to reconnect right?
If so i personally would demote this here to a Warning, as well as the default case here.

Then on a complete error we will in the if loop there return the error which than get logged outside of this function as an error and throws the reconnection error right?

beckhoff_ads_plugin/connect.go
```diff
@@ -0,0 +1,122 @@
...
+	for len(needed) > 0 {
+		select {
+		case update := <-g.NotificationChan:
+			if update != nil {
+				g.pendingInitial = append(g.pendingInitial, update)
+				delete(needed, strings.ToLower(sampleName(update)))
+			}
+		case <-initialCtx.Done():
+			g.Log.Warnf("Timed out waiting for initial samples; %d symbols not yet received: %v", len(needed), needed)
+			return nil
```
led0nk, line 136: won't this happen from a bad connection or so? doesn't it then make sense to return the initialCtx.Err() here as well if so?

beckhoff_ads_plugin/logging.go
led0nk: unresolving this just so that you see this message:

for the future what we want to have is:

- the library doesn't do the logging for us
- the library supports all methods and always returns specific errors
- based on the errors, we will do the log-handling within benthos-umh
- this will then make this here redundant and we can get rid of this

led0nk (changes requested):

beckhoff_ads_plugin/client_goads.go
RuneRoven: this is mainly if you run standalone benthos and put it in a container. Using it in the umh-core it will always be a container so it depends on how you think about it. I tried to find ways to check what the actual host ip was or find out if running inside a container to throw a warning based on that but i could not find any good solution. If running standalone benthos you can derive the hostAMS from the actual host ip but inside a container you will always need to specify the hostIP and derive the hostAMS from that. For umh-core we could change the hostIP input to required but then you will need to specify this when running standalone which i don't like. We can for sure remove it and enhance the documentation about this.

beckhoff_ads_plugin/ads.go
RuneRoven: You are correct, i removed it.

beckhoff_ads_plugin/client_goads.go
RuneRoven: I leave it for now but i remove the 10.x.x.x and made a range array with adrresses to check instead so it is easy to change. Could remove the entire thing if we think that is is the best solution

beckhoff_ads_plugin/client_goads.go
RuneRoven: Yes, this is used in the route creation. I will run an additional test again on this to validate this. The route creation and response id/ip is not well documented so much of this is reverse engineered. It might work with container ip as the netid.

RuneRoven: I have done some more testing now and updated my internal documents about this. It seems to work from container ip in TC3 but not in TC2. So for now we will leave this and the isLikelyContainerIP. It looks like TC2 uses the hostIP as unique identifier and TC3 uses the netID. I will later deep dive in this when i implement the secure ADS.

beckhoff_ads_plugin/connect.go
RuneRoven: Yes, if all fails to register we error out and reconnect. Changed this to a warning and i group the fails to and array and then loop the warnings.

beckhoff_ads_plugin/connect.go
RuneRoven: a  registered symbol always delivers a first sample if everything is correct.
Did some more testing for this, and found an edge case with the  fixed 10s wait against a user-configurable `maxDelay`. With `maxDelay: 20s`: warning at 10.0s, first sample at 14.7s — the PLC may hold a notification for up to cycleTime + maxDelay, so the gate expired before delivery was due. Now derived from the slowest symbol (`cycleTime + maxDelay + initialSampleWait`);  same config, no timeout. Defaults go 10s -> 10.2s.

A bad connection will not get here unnoticed.  `client.Connect` or `AddNotifications` would have failed first, and one that dies during the wait is caught by `IsClosed()` -> `ErrNotConnected` on the next read. So lets keep the `nil`. Erroring would re-register every second forever  and l block the other symbols.

beckhoff_ads_plugin/client_goads.go
led0nk: honestly i think removing this is the best way here, because one could have a very weird docker networking setup and we will never be able to predict this.

Then i think we should not at any circumstance "restrict" this and even check for this.

I'd personally stay with this configoption but not this check here

beckhoff_ads_plugin/client_goads.go
led0nk: this IMO is more evidence on the targetIP/Port thing which gets parsed by a helper function in the first place from my perspective.

But other than that i think we could make things "more clear", which also gets back to the config.
So having something that basically states us here "tc2/tc3" then one would only need to hand over the address in this function here as well, because whenever this here is called you could have a switch running on which type of the protocol were using here.

Not a hard opinion, but reduces a bit complexity and makes the code more readable in regards of "ah this is the tc2 path"

beckhoff_ads_plugin/logging.go
led0nk: btw, forgot about this - this is maybe stuff for the rollout-ticket

led0nk (changes requested):
looks good to me, but would prefer to solve the things that are still open (the logging thing is just to take note on the ticket for future plans)

PR #402, opened 2026-07-01
feat(ads): Beckhoff ADS plugin — parsing (4/6)

RuneRoven:
Type conversion / metadata (BOOL, INT, REAL, STRING, ARRAY): `data_type`/`base_type`/`data_size`, symbol-cache population, and the individual-read fallback.

---


RuneRoven:
@coderabbitai review

PR #403, opened 2026-07-01
test(ads): Beckhoff ADS plugin — unit tests (2/3)

RuneRoven:
Unit tests + testdata (Ginkgo). PLC/hardware specs are gated by `TEST_ADS_*` env and skip without hardware. Verified locally: 79 passed / 66 skipped / 0 failed.

---


RuneRoven:
@coderabbitai review

beckhoff_ads_plugin/deterministic_helpers_unit_test.go
```diff
@@ -0,0 +1,1921 @@
...
+targetAddress: "1.2.3.4"
+`, nil)
+			Expect(err).NotTo(HaveOccurred())
+
+			_, err = NewAdsCommInput(conf, service.MockResources())
+			Expect(err).To(MatchError(ContainSubstring("unifiedAddress is required")))
+		})
+	})
+
+	Describe("adsValueBytes", func() {
```
FalconTube, line 461: This test here should also cover this section in `parse.go`

```golang
if err != nil {
        b = []byte(strconv.Quote(decoded))
}
```

beckhoff_ads_plugin/deterministic_helpers_unit_test.go
```diff
@@ -0,0 +1,1921 @@
...
+			Expect(rotateLeft64(1, 0)).To(Equal(uint64(1)))
+			Expect(rotateLeft64(1, 1)).To(Equal(uint64(2)))
+			Expect(rotateLeft64(1, 63)).To(Equal(uint64(1 << 63)))
+			Expect(rotateLeft64(1, 64)).To(Equal(uint64(1))) // wraps
+		})
+	})
+})
+
+var _ = Describe("Plugin Internal Functions", func() {
+	Describe("CreateSymbolList", func() {
```
FalconTube, line 305: This test section mostly covers the happy path only.
The following `!ok / err != nil` clauses of this section of `config.go` are not covered:

```
target, ok := targets[key]
if !ok {
        warnings = append(warnings, fmt.Sprintf("symbol %q: ignoring unknown option %q (supported: maxDelay, cycleTime)", fullSpec, key))
        continue
}
d, err := parseSymbolDuration(value)
if err != nil {
        warnings = append(warnings, fmt.Sprintf("symbol %q: ignoring invalid %s value %q (using default)", fullSpec, key, value))
        continue
}
```

FalconTube (changes requested):
Also pls add tests for these functions in `connect.go`:

1. `initialSampleTimeout`
2. `waitForInitialSamples`<br>

Big parts of `client_goads.go` are not covered by tests. I think it is like this, because it would require more heavy mocking and we should solve this via integration tests. Is that correct?
Same goes for the `ads.go` file, where `Connect` and `Close` are not covered. Probably also better done via integration tests

beckhoff_ads_plugin/deterministic_helpers_unit_test.go
```diff
@@ -0,0 +1,1870 @@
...
+		It("JSON-quotes a string containing special characters correctly", func() {
+			payload, tagType := adsValueBytes("STRING", `has "quotes"`)
+			Expect(tagType).To(Equal("string"))
+			var roundTrip string
+			Expect(json.Unmarshal(payload, &roundTrip)).To(Succeed())
+			Expect(roundTrip).To(Equal(`has "quotes"`))
+		})
+	})
+
+	Describe("ReadBatchNotification initial-sample flush", func() {
```
FalconTube, line 487: This test suite here does not cover the entire bottom part of `ReadBatchNotification` in `read.go` → Everything after the `context` usage.<br><br>We should cover it, since it is very crucial part of software.
Also, can use existing setup without a lot of heavy mocking

beckhoff_ads_plugin/deterministic_helpers_unit_test.go
RuneRoven: added what is meaningful and On client_goads.go: yes for the nine pass-throughs (Connect, Close, IsClosed, LoadSymbols, GetSymbol, ReadMultipleSymbols, ReadFromSymbol, AddNotifications,
  ▎ newGoADSClient). They're 1:1 calls into *adsLib.Session so yes, this will require heavy mocking and is today covered by hardware tests.

RuneRoven: While covering waitForInitialSamples I checked the first-sample behaviour against hardware, because the docs claimed the PLC samples immediately at registration. It does
  ▎ not — the first sample lands at the first change-check, one cycleTime later (measured 11ms at cycleTime: 0, 120ms at 100ms, 60.02s at 60s, on TC2 2.10.1320 and TC3
  ▎ 3.1.4026, confirmed at the wire). That means initialSampleTimeout's per-symbol cycleTime + maxDelay term is load-bearing rather than defensive, and the docs page is
  ▎ corrected in #404.

FalconTube (approved):
Great improvements, good to merge now!

PR #409, opened 2026-07-08
[ENG-5086] Remove name from unified address

rpaland:
## What
The name part of the unifiedAddress was mainly a duplication of the tag name column and potentially caused confusion for customers. Therefor the name was strapped from the unified address. Though, for backwards compatibility, the old format still works. In this case the name part of the unified address is parsed into the tag name column and on next save and deploy the config is adjusted.


rpaland:
Further product decision to be made

PR #414, opened 2026-07-13
fix: schema validation error is misleading, now checks on datatype

led0nk:
# Description

Whenever the user has set up a datamodel `typevalidation_v1` with e.g.:

```yaml
count:
    _payloadshape: timeseries-number
```

If he then wants to write a string into e.g. `enterprise.line.site._typevalidation_v1.count` he gets no clear guidance by the errormessage, as you could see from the ticket.
It simply lists the available tags and states that the `virtual_path` is wrong.

Therefore we want and introduce here a type-mismatch-error message, which clearly guides for this example the user to read something like: `want: timeseries-number` `got: timeseries-string`. 
With this one doesn't have to think about each individual tag, as the list of tags that get printed back from the original error-messages grows whenever the datamodel grows.

This PR introduces a clear type-mismatch-error-message.

This is tested within the [datacontract-skip-PR](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2642) and looks like this now:

grafik

Fixes ENG-5319


JeremyTheocharis (approved):

rpaland (approved):
LGTM

PR #416, opened 2026-07-15
fix(sparkplug): use per-metric timestamp for spb_timestamp (ENG-5341)

Aaron99B:
## Customer problem

A customer saw Sparkplug metrics go missing between the device and the UNS. Values published at the edge never arrived downstream.

## Background

The `sparkplug_b` input splits each Sparkplug payload into one message per metric and gives each message a timestamp. Sparkplug carries two timestamps: one on the payload for when the message was sent, and one on each metric for when that value was sampled. The input used the payload timestamp for every metric and ignored the per-metric one.

When a device sends the same tag more than once in a payload, for example buffered samples flushed after a reconnect, every copy got the same timestamp. The downsampler drops a sample whose timestamp matches the last one it kept for that tag, so the copies collapsed to a single point and the rest were dropped.

## Fix

Use the metric's own timestamp when it is set, and fall back to the payload timestamp only when it is missing (in `createMessageFromMetric` and the UMH-conversion path). Each sample now carries the time it was captured, which is what the Sparkplug B spec defines the per-metric timestamp to mean, so the downsampler keeps samples whose timestamps differ.

Testing: a unit test covers distinct and fallback timestamps; an integration test publishes a two-metric NDATA through a real broker and checks the timestamps stay distinct.

Fixes [ENG-5341](https://linear.app/united-manufacturing-hub/issue/ENG-5341/sparkplug-b-input-use-per-metric-timestamp-metrictimestamp-instead-of)

### DIFF
--- a/sparkplug_plugin/sparkplug_b_input.go
+++ b/sparkplug_plugin/sparkplug_b_input.go
@@ -1138,8 +1138,14 @@ func (s *sparkplugInput) createMessageFromMetric(metric *sparkplugb.Payload_Metr
 	msg.MetaSet("spb_metric_index", fmt.Sprintf("%d", metricIndex))
 	msg.MetaSet("spb_metrics_in_payload", fmt.Sprintf("%d", totalMetrics))
 
-	if payload.Timestamp != nil {
-		msg.MetaSet("spb_timestamp", fmt.Sprintf("%d", *payload.Timestamp))
+	// ENG-5341: prefer the per-metric timestamp (Sparkplug proto field 3) so each metric
+	// keeps its own sample time; fall back to the payload-level timestamp when absent.
+	ts := payload.Timestamp
+	if metric.Timestamp != nil {
+		ts = metric.Timestamp
+	}
+	if ts != nil {
+		msg.MetaSet("spb_timestamp", fmt.Sprintf("%d", *ts))
 	}
 
 	// Add metric-specific metadata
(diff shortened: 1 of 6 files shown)


sparkplug_plugin/sparkplug_b_input.go
```diff
@@ -1605,8 +1611,10 @@ func (s *sparkplugInput) tryAddUMHMetadata(msg *service.Message, metric *sparkpl
 		sparkplugMsg.MetricName = "unknown_metric"
 	}
 
-	// Set timestamp from payload if available
-	if payload.Timestamp != nil {
+	// ENG-5341: prefer the per-metric timestamp, falling back to the payload-level one.
+	if metric.Timestamp != nil {
```
led0nk: don't really like those if else loops, can we try to make use of either a helper-function or the pattern you already used above with setting a variable first?

led0nk (changes requested):

sparkplug_plugin/sparkplug_b_input.go
```diff
@@ -1138,8 +1138,14 @@ func (s *sparkplugInput) createMessageFromMetric(metric *sparkplugb.Payload_Metr
...
-	if payload.Timestamp != nil {
-		msg.MetaSet("spb_timestamp", fmt.Sprintf("%d", *payload.Timestamp))
+	// ENG-5341: prefer the per-metric timestamp (Sparkplug proto field 3) so each metric
+	// keeps its own sample time; fall back to the payload-level timestamp when absent.
+	ts := payload.Timestamp
+	if metric.Timestamp != nil {
+		ts = metric.Timestamp
+	}
+	if ts != nil {
+		msg.MetaSet("spb_timestamp", fmt.Sprintf("%d", *ts))
```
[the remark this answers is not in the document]
Aaron99B: changed the documentation

sparkplug_plugin/sparkplug_b_input.go
Aaron99B: Moved it to a helper function

led0nk (approved):

PR #418, opened 2026-07-20
fix(historian): don't warn or degrade on contract-mismatch-only batches

Aaron99B:
## What

A TimescaleDB historian stores exactly one data contract but usually reads from a broader UNS subscription, so it drops messages for other contracts by design. How those drops were logged caused two problems:

1. A fully-dropped batch was logged with `Warnf`, which umh-core can treat as unhealthy, so a historian that received an all-other-contract batch reported degraded even though nothing was wrong.
2. Even once that was quieted, the logs could not separate two very different situations:
   - Wrong `data_contract_name` or subscription: nothing is ever stored (broken).
   - Over-broad subscription: the contract is stored, alongside expected drops (fine).

"Is any data reaching the historian at all?" is a cumulative question and could not be answered from any single batch.

## Change

Logging now keys on whether a value row for the configured contract has ever been stored:

- First successful store: info, once. `first message stored for data contract _<contract>`.
- Receiving data but nothing stored yet, and a batch is entirely other-contract: info, once. `nothing stored for data contract _<contract> yet; the last N message(s) were all for other contracts. Check data_contract_name and the topics this flow subscribes to`.
- Entirely-other-contract batch after data has flowed: debug (an expected lull).
- Genuine fault (matching contract, missing or unusable value, write failure): warn, unchanged.

Every contract-mismatch path is info or debug, so the bridge stays healthy; only genuine faults warn. Messages and comments use standalone benthos-umh wording, since this output can run outside umh-core.

## Behavior for the two cases

- Wrong contract: only `nothing stored for _<contract> yet` appears. No data is landing, and the message names the two things to check.
- Over-broad subscription: `first message stored` appears once, then other-contract batches are quiet. Data is confirmed flowing.

## Tests

Integration specs against a live TimescaleDB container cover:

- never stored: `nothing stored ... yet` logged once at info, no warning or error;
- stored, then an other-contract-only batch: `first message stored` once, no starving notice, no warning;
- genuine fault: still warns.

The test log capture records benthos's `level=... msg=...` format, so the specs assert severity, not only message text.

## Note

A contract mismatch, including a `data_contract_name` typo that matches nothing, is logged at info and never marks the bridge degraded. This is deliberate: the case stays visible in the logs without turning a bridge red.

Closes ENG-5394

### DIFF
--- a/.changelog/unreleased/fixes-20260720-132909.yaml
+++ b/.changelog/unreleased/fixes-20260720-132909.yaml
@@ -1,3 +1,3 @@
 kind: fixes
-body: Historian no longer logs a warning (or reports the bridge as degraded) when a batch is dropped solely because its messages belong to a different data contract than the one the bridge stores; that expected case is now a debug-level message (ENG-5394)
+body: The TimescaleDB historian output no longer logs a warning when a batch is dropped solely because its messages belong to a data contract other than the one it stores; that expected case is now a single info-level notice per process, while genuine faults are still logged as warnings (ENG-5394)
 time: 2026-07-20T13:29:09.837735+02:00
--- a/.changelog/unreleased/fixes-20260720-132909.yaml
+++ b/.changelog/unreleased/fixes-20260720-132909.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: Historian no longer logs a warning (or reports the bridge as degraded) when a batch is dropped solely because its messages belong to a different data contract than the one the bridge stores; that expected case is now a debug-level message (ENG-5394)
+time: 2026-07-20T13:29:09.837735+02:00
--- a/.changelog/unreleased/fixes-20260720-132909.yaml
+++ b/.changelog/unreleased/fixes-20260720-132909.yaml
@@ -1,3 +1,3 @@
 kind: fixes
-body: The TimescaleDB historian output no longer logs a warning when a batch is dropped solely because its messages belong to a data contract other than the one it stores; that expected case is now a single info-level notice per process, while genuine faults are still logged as warnings (ENG-5394)
+body: The TimescaleDB historian output logs a one-time info notice confirming the first message stored for its data contract. When it is receiving data but has stored nothing for that contract, it logs a one-time notice that the messages all belong to other contracts and points at data_contract_name and the flow's topic subscription. Other-contract-only batches no longer warn once data is flowing, while genuine faults are still logged as warnings (ENG-5394)
 time: 2026-07-20T13:29:09.837735+02:00
--- a/historian_plugin/output_historian.go
+++ b/historian_plugin/output_historian.go
@@ -43,7 +43,7 @@ func historianConfig() *service.ConfigSpec {
 		Field(service.NewStringField("database").Description("Database name.").Default("umh")).
 		Field(service.NewStringField("username").Description("Login role.").Default("umh_owner")).
 		Field(service.NewStringField("sslmode").Description("require | disable | verify-full.").Default("require").Examples("require", "disable", "verify-full")).
-		Field(service.NewStringField("sslrootcert").Description("CA cert path (inside the umh-core container).").Default("").Advanced()).
+		Field(service.NewStringField("sslrootcert").Description("CA cert path, as seen by the benthos process.").Default("").Advanced()).
 		Field(service.NewStringField("sslcert").Description("Client cert path.").Default("").Advanced()).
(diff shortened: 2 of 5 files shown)


historian_plugin/output_historian.go
```diff
@@ -444,7 +448,15 @@ func (o *historianOutput) WriteBatch(ctx context.Context, batch service.MessageB
 		// A fully-dropped batch writes nothing while the connection stays up, so umh-core would
 		// see a healthy bridge silently discarding data. Warn (umh-core surfaces this as
 		// degraded); still return nil so one bad message never stalls the stream.
-		if len(batch) > 0 {
+		switch {
+		case len(batch) == 0:
+			// empty batch: nothing to say
+		case contractMismatches == len(batch):
+			// Every message simply belonged to another data contract. Expected when a bridge
```
led0nk: can we please not use umh-core related words here, since this should also work for benthos-umh as it's a separate appliucation. Also i can see that this is used in other comments here. Might make sense to also replace this and or make thosse comments less verbose as well.

historian_plugin/output_historian.go
```diff
@@ -444,7 +448,15 @@ func (o *historianOutput) WriteBatch(ctx context.Context, batch service.MessageB
...
+			// subscribes to more of umh.v1 than the single contract this historian stores, so
+			// it is not a fault: log at debug and leave the bridge healthy.
+			o.logger.Debugf("TimescaleDB historian: dropped all %d message(s) in the batch; all belong to other data contracts (this historian stores only _%s). This is expected when the bridge subscribes to more than _%s and is not an error.", len(batch), o.contract, o.contract)
```
led0nk: Same then applies for the error-messages here / debug-messages. shouldn't refer to a `bridge´ here.

On top of that i'm wondering if this is really debug or should be promoted to info, can talk about this (which is kinda the whole point of this PR)

.changelog/unreleased/fixes-20260720-132909.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: Historian no longer logs a warning (or reports the bridge as degraded) when a batch is dropped solely because its messages belong to a different data contract than the one the bridge stores; that expected case is now a debug-level message (ENG-5394)
```
led0nk: same applies here the description should be matching for benthos-umh as a "standalone"-application this can then be changed, whenever this arrives in umh-core, but here it should be a single working plugin not related to umh-core directly

led0nk (changes requested):

.changelog/unreleased/fixes-20260720-132909.yaml
Aaron99B: changed that

historian_plugin/output_historian.go
Aaron99B: done

historian_plugin/output_historian.go
Aaron99B: done

Aaron99B: Changed it to be promoted to Info

historian_plugin/output_historian.go
```diff
@@ -441,10 +449,26 @@ func (o *historianOutput) WriteBatch(ctx context.Context, batch service.MessageB
...
 	o.warnHighChurnMetadata(churn)
 	if len(rows) == 0 {
-		// A fully-dropped batch writes nothing while the connection stays up, so umh-core would
-		// see a healthy bridge silently discarding data. Warn (umh-core surfaces this as
-		// degraded); still return nil so one bad message never stalls the stream.
-		if len(batch) > 0 {
+		// Return nil in every branch so one fully-dropped batch never stalls the stream.
+		switch {
+		case len(batch) == 0:
+		case contractMismatches == len(batch):
```
led0nk:
```go
 if len(rows) == 0 {
                // Return nil in every branch so one fully-dropped batch never stalls the stream.
                switch {
                case len(batch) == 0:                                   
                case contractMismatches < len(batch):
                        o.logger.Warnf(...)                                  
                case o.everStored.Load():
                        o.logger.Debugf(...)                                 
                default:
                        o.noteStarving(len(batch))                          
                }
                return nil
        }
```

think we could make use of something like this here with noteStarving being a helper-function. With this we reduce the amount of nesting from if -> switch -> if else loops

led0nk (changes requested):

historian_plugin/output_historian.go
Aaron99B: Did that

led0nk (approved):

PR #419, opened 2026-07-23
fix: tag_processor string conversion without number

led0nk:
# Description

When reading data with e.g. `opcua_plugin` via bridges it occured, that without setting `msg.meta.datatype = string` that a number in a string got autoconverted to a number. When in fact this has been a string originally and should end up as a string.

Therefore this PR should fix this issue and not convert any numbers in strings anymore. This way it should then also get rid of the scientific annotation.

As follow-ups we should ensure that our input-plugins also directly support this. Meaning they parse it from the beginning in the correct format. Currently this does not happen for e.g. `s7comm`, `modbus`, `ethernetip`, `sensorconnect`. 

This then is more about aligning, that each plugin works the same.

### DIFF
--- a/.changelog/unreleased/fixes-20260723-120000.yaml
+++ b/.changelog/unreleased/fixes-20260723-120000.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: The tag processor no longer converts numeric-looking string values to numbers; a string read from the source stays a string in the UNS, and msg.meta.datatype = "number" remains available for explicit coercion (ENG-5422)
+time: 2026-07-23T17:31:13+02:00
--- a/tag_processor_plugin/conversion_test.go
+++ b/tag_processor_plugin/conversion_test.go
@@ -1,0 +1,55 @@
+// Copyright 2026 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tag_processor_plugin
+
+import (
+	"encoding/json"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = DescribeTable("convertValue datatype handling",
+	func(input any, datatype string, want any, wantErr bool) {
+		p := &TagProcessor{}
+		got, err := p.convertValue(input, datatype)
+		if wantErr {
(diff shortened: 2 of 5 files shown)


tag_processor_plugin/conversion_test.go
```diff
@@ -0,0 +1,55 @@
...
+	// auto-detect (no msg.meta.datatype set)
+	Entry("bool passes through", true, "", true, false),
+	Entry("float to json.Number", 23.5, "", json.Number("23.5"), false),
+	Entry("int to json.Number", 42, "", json.Number("42"), false),
+	Entry("non-numeric string stays string", "hello", "", "hello", false),
+	Entry("numeric string stays string", "2340925", "", "2340925", false),
+	Entry("float-string stays string", "23.5", "", "23.5", false),
+	Entry("large-int string stays exact", "12345678901234567890", "", "12345678901234567890", false),
+	Entry("array to JSON string", []any{1, 2}, "", "[1,2]", false),
+	Entry("object to JSON string", map[string]any{"a": 1}, "", `{"a":1}`, false),
```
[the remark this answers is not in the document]
led0nk: noone cares about claude.md

.changelog/unreleased/fixes-20260723-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: The tag processor no longer converts numeric-looking string values to numbers; a string read from the source stays a string in the UNS, and msg.meta.datatype = "number" remains available for explicit coercion (ENG-5422)
```
FalconTube, line 2: Pls remove `ENG-5422` slug :)

tag_processor_plugin/tag_processor_plugin.go
```diff
@@ -627,9 +623,9 @@ func (p *TagProcessor) autoConvertValue(v any) any {
 		return string(jsonBytes)
 	default:
 		str := fmt.Sprintf("%v", val)
-		num, err := strconv.ParseFloat(str, 64)
+		_, err := strconv.ParseFloat(str, 64)
 		if err == nil {
-			return json.Number(fmt.Sprintf("%v", num))
+			return json.Number(str)
 		}
```
FalconTube, line 631: This is probably personal preference, but I never really do/see in code a check for `if err == nil `.
Would change it to `if err != nil` and then revert the return logic.

I was first wondering if this was the correct logic until I saw that you check for the `==` instead of `!=` :D 
Might trip up other gophers as well.

FalconTube (changes requested):
Code looks good, only minor comments.
Checked it in product as well, works great!

But still might make sense to bump next benthos version to `v0.14.0`, since this here introduces a breaking behavior, although we are fine with it in `umh-core`

.changelog/unreleased/fixes-20260723-120000.yaml
led0nk: unfortunately this is meant to be when coming from benthos-umh and should be stripped away when the PR goes to umh-core 😄

FalconTube: Ahhh okay got it :D

FalconTube (approved):
Looks good now!

PR #425, opened 2026-07-30
feat(historian): surface every discarded message and guard unvalidated inserts

Aaron99B:
## Problem

Six historian tickets with one root cause: dropped messages went to a debug log, so a bridge writing nothing looked identical to a healthy one. Separately, a bridge subscribing to more of the UNS than it writes reported the throughput it *accepted* (12,000/s) rather than the rows it stored (84/s), because that figure is `output_sent`.

## Behaviour matrix

Every case the output can hit, and what it does now. "Batch" is the benthos ACK/NACK signal; "rest of batch" means the other messages are still written.

| Case | Ticket | `reason=` | Batch | Rows | Log |
|---|---|---|---|---|---|
| Any message carries another data contract | ENG-5413 | `contract_mismatch` | **NACK**, whole batch | none, not even the matching messages | error, throttled to 1 per 2 min, plus the `uns` input's own error per batch |
| Versioned contract whose schema was bypassed | ENG-5419 | `contract_bypassed` | ACK | none for that message, rest of batch written | error, 1 line per batch |
| Unversioned contract (`_historian`) | ENG-5419 | — | ACK | stored | none |
| Tag changes datatype, `allow_datatype_changes` unset | ENG-5468 | poison, counted on `historian_rows_poisoned{phase=resolve,sqlstate=P0001}` | ACK | offending row only, rest of batch written | error per row, naming the flag that would store it |
| Tag changes datatype, `allow_datatype_changes: true` | ENG-5468 | — | ACK | stored, tag holds both types | info once, on the first stored message |
| Payload carries a field beyond `value`/`timestamp_ms` | ENG-5415 | `not_timeseries` | ACK | none for that message, rest of batch written | error, 1 line per batch |
| `value` absent or null | ENG-5412 | `missing_value` | ACK | none for that message, rest of batch written | error, 1 line per batch, with the fix |
| `timestamp_ms` absent or null | ENG-5412 | `missing_timestamp` | ACK | none for that message, rest of batch written | error, 1 line per batch, with the fix |
| Payload is not JSON, or not a JSON object | ENG-5414 | `not_structured` / `not_object` | ACK | none for that message, rest of batch written | error, 1 line per batch |
| `umh_topic` absent or unparseable | ENG-5414 | `invalid_topic` | ACK | none for that message, rest of batch written | error, 1 line per batch |
| Non-finite number, or unparseable timestamp | ENG-5414 | `unclassifiable_value` / `bad_timestamp` | ACK | none for that message, rest of batch written | error, 1 line per batch |
| Virtual path under `Root.Objects.Server` | — | `server_virtual_path` | ACK | none for that message, rest of batch written | **debug**, 1 line per batch |
| Same `(tag, ts)` with a different value | — | poison, counted on `historian_rows_poisoned{phase=value}` | ACK | offending row only, rest of batch written | error per row |
| Standing write fault: missing privilege, disk full, unknown SQLSTATE | — | — | **NACK**, held for retry | none | error, loudly, every attempt |
| Transient write fault: connection loss, deadlock, lock wait, timeout | — | — | **NACK**, held for retry | none | quiet, benthos retries |

Three properties of the table worth stating outright:

* **Only two things ever NACK**: a foreign data contract, and a write fault. Everything else ACKs, so one bad message never stalls the stream.
* **Every drop increments `messages_dropped{reason=…}`**, including the NACKed mismatch, so the metric is a complete account even where the log is throttled. The two poison cases are the exception: they are counted on `historian_rows_poisoned`, because they fail at the database rather than in the transform.
* **Nothing is held back at startup.** Every reason, mismatch included, logs and counts from the first batch.

## What we are shipping

* **Drops log at error, one aggregated line per reason per batch** (ENG-5414, ENG-5412), carrying the count, an example topic and that reason's fix. `missing_value_or_timestamp` splits into `missing_value` and `missing_timestamp`. Per-message logging was untenable once the config-level reasons arrived: a reason that fires on every message produced 12,000 identical lines per second on a misconfigured bridge. `server_virtual_path` stays at debug, since a broad OPC UA browse picking up server diagnostics is not something the operator did wrong.

* **A batch carrying any foreign contract is refused** (ENG-5413). `WriteBatch` errors before writing anything, so benthos never adds it to `output_sent` and the Management Console figure becomes rows stored. This reverses [ENG-5394](https://linear.app/united-manufacturing-hub/issue/ENG-5394/historian-contract-mismatch-only-batches-log-a-misleading-dropped-all), which made the case quiet on the premise that a broad subscription is legitimate; the premise is what changed. The inverted specs in `output_integration_test.go` are the record of that reversal and the place to start reviewing.

  The error carries everything the fix needs: the configured contract, how much of the batch was foreign, **the contracts that arrived instead**, and the `umh_topics` pattern to paste:

  ```
  TimescaleDB historian: no message carries data contract _drops, 1 of 1 message(s)
  carry [_other_v1] instead (reason=contract_mismatch). Either set data_contract_name
  to a published contract, or narrow umh_topics to ^umh\.v1(?:\.[^._][^.]*)+\._drops(_v\d+)?\..+$
  ```

  The operator's next action is to narrow `umh_topics`, and they cannot write that pattern without knowing what arrived. Two bounds: the set stops growing one entry past a cap of five, so a subscription selecting hundreds of contracts neither parses every topic in a batch it is about to discard nor lists them all, rendering `[_a, _b, _c, _d, _e, and others]`. Sorted, so repeated reports match instead of reshuffling with map order.

* **A datatype change on a tag needs an opt-in** (ENG-5468). `allow_datatype_changes` (default `false`) is the whole of it:

  ```
  no datatype change                    -> stored
  datatype change, flag unset           -> offending rows dropped as poison
  datatype change, allow_datatype_changes: true -> stored, tag holds both types
  ```

  It reads the same on every contract, versioned or not. With it set, the tag resolves on its natural key: one `topic_id` holds numeric rows in `value_num` and text rows in `value_text`, and `umh.tag.value_type` keeps only the first type seen, so reads need `coalesce(value_num::text, value_text)`.

  Without it, the drop already worked before this branch: the type-keyed lookup misses, the guarded upsert raises, the row is dropped and the rest of the batch is written. What is new is that the log names the flag. `phase=resolve` with `sqlstate=P0001` can only be the datatype guard: the other sanctioned P0001 source, `raise_pk_conflict`, runs on the value and attribute inserts. No string-matching on the trigger's message.

* **Unvalidated data is stored; a broken schema is not** (ENG-5419). A **versioned** contract carrying `data_contract_bypassed=true` is dropped as `contract_bypassed`, and no setting overrides it: the version names a schema that was never checked. An **unversioned** contract is stored, because it never named one.

  Worth a look in review: the version check in front of that meta is not a narrowing, and now carries a comment saying why. The `uns` output sets `data_contract_bypassed=true` on **every** unversioned message (`uns_plugin/schema_validation/validator.go:198-210`), so reading the meta on its own would drop all `_historian` traffic under a reason describing a different fault.

* **A payload of exactly `{value, timestamp_ms}`** (ENG-5415). A work order that happens to include both is rejected as `not_timeseries` rather than narrowed to those two fields.

* **`historian_messages_dropped` becomes `messages_dropped`**, the name the other five plugins already use, so [united-manufacturing-hub#2681](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2681) can subtract output-path drops from `output_sent` with one metric case instead of a per-plugin one ([ENG-5507](https://linear.app/united-manufacturing-hub/issue/ENG-5507/umh-core-exclude-dropped-messages-from-write-flow-throughput)).

* **A data-loss bug that predates this branch** (`b9e3341`): the `Root.Objects.Server` filter was a `HasPrefix`, so a real tag under `Root.Objects.ServerRoom` was discarded as OPC UA diagnostics. Only the true positive had a spec, and at debug level the loss was silent.

## What we are NOT shipping

* **No partial credit on a refused batch.** Benthos gives an output no way to reduce `output_sent` for part of a batch, and per-message nacking is not an option either. A NACK does not preserve the data: the `uns` input leaves offsets uncommitted (`uns_input.go:173-189`) and the next ACKed batch commits past them. So a bridge with a wrong subscription stores nothing and loses the mismatched traffic, which is the pressure to fix the subscription. Worth reviewing as a deliberate cost.

* **The 30s startup grace is gone, not replaced.** It was a umh-core-specific constant in a plugin that also has to work under plain benthos, where it only withheld real errors while data was being discarded. So a bridge whose source emits malformed data will now not reach Active either, rather than coming up running-but-red. The durable fix is a starting-state timeout in umh-core.

* **No schema work for the mixed-datatype case**: no trigger change, no new table, no migration. `umh.tag.value_type` keeps reporting the first type seen even once the tag holds both, so it stops describing every row. That is a documented cost of the flag, not an oversight.

* **The historian no longer refuses unvalidated data, so ENG-5419 does not ship as written.** Its premise was that only schema-validated data is stored. What ships is narrower: data that *claims* a schema and did not get one is refused, data that never claimed one is stored. The trade was deliberate: refusing `_historian` outright meant every default bridge needed a config field before it could store a single row, to guard against a datatype change that most tags never make.

* **Nothing verified against a real umh-core FSM.** The suite pins the NACK and the log line, but not whether a bridge erroring on its first batch sticks in starting, rolls back, or comes up red, and not whether `IsLogsFine` matches the emitted format. One manual run is the main thing to check before merging.

* No UI or frontend changes.

## Before this can be shipped

* **Both ManagementConsole templates need their subscription narrowed**, and no longer a new config field. They use `topics: ['^umh\.v1\..*']` with `data_contract_name: historian`, so after this lands a default Historian bridge refuses every batch and never reaches Active. They need:

  ```yaml
  topics:
    - '^umh\.v1(?:\.[^._][^.]*)+\._historian(_v\d+)?\..+$'
  ```

  The pattern follows `uns_topic.go:279-318`. The `(?:\.[^.]+){0,8}` form in the current template comment is wrong twice: it caps location depth at 8, and its `[^.]+` also matches a virtual-path segment named `_historian`. No `allow_datatype_changes` is required: a Historian bridge stores on the default, and the flag is only wanted once a tag actually changes type.

* **Existing customer bridges are data-affecting, not cosmetic.** They carry the broad subscription in `config.yaml`, stop storing after the bump until narrowed, and lose the mismatched messages they were previously discarding. Exposure looks small (the output shipped in v0.13.2, current pin is 0.14.0, both templates sit behind the `historian` flag), so confirm the install base before release.

* Anything scraping `historian_messages_dropped` or the `missing_value_or_timestamp` label needs updating. Nothing in umh-core or the Management Console reads them today.

## Testing

176 unit specs, plus 49 that need a live TimescaleDB.

The integration suite first passed *without any change*, which was the warning sign: the existing specs asserted only the `reason=…` substring, which matches at any log level, so the Debugf-to-Errorf change had no coverage. Added assertions pinning `level=error`, then deleted the exclusion and the guards to confirm the specs fail.

Every spec asserting a drop was 1-of-1, so nothing proved the per-batch aggregation aggregates. There is now a four-message batch (one good, two `missing_value`, one `missing_timestamp`) asserting `dropped 2 of 4` and `dropped 1 of 4` on separate lines, exactly two error lines, and one row still written. Before it, a regression back to per-message logging would have passed the whole suite.

The datatype-flip hint has both a unit spec on the discriminator and integration specs on each side of it: a flip logs the flag, an append-only conflict at the same `(tag, ts)` does not, since the flag cannot fix that one. Negative control: inverting the guard fails exactly those specs and nothing else.

The arrived-contract rendering has unit specs for its sorting, its cap and its overflow form, plus one for a topic that does not parse being skipped rather than reported as a blank contract. Its negative control neuters only the join, which fails those three plus the integration spec at `output_integration_test.go:880` and nothing else.

Fixes ENG-5412, ENG-5413, ENG-5414, ENG-5415, ENG-5419, ENG-5468


historian_plugin/transform.go
```diff
@@ -188,49 +188,81 @@ type Row struct {
...
+	DropMissingTimestamp    DropReason = "missing_timestamp"
+	DropUnclassifiableValue DropReason = "unclassifiable_value"
+	DropBadTimestamp        DropReason = "bad_timestamp"
+	DropContractBypassed    DropReason = "contract_bypassed"
+	DropContractUnvalidated DropReason = "contract_unvalidated"
+	DropNotTimeseries       DropReason = "not_timeseries"
 )
 
+var dropHints = map[DropReason]string{
+	DropMissingValue:        " -- the historian needs a {value, timestamp_ms} payload; set msg.payload.value in the processing step",
```
led0nk: is this correct here? one can change / adjust the payload in the processing but generally speaking this should come from the input_plugin, which get's handed over to the processor.

Also this here is referring to `nodered_js` and `tag_processor` how does this work then for e.g. bloblang? 
Generally speaking this should just refer to a missing value in the payload, without giving instructions (too detailed)

historian_plugin/transform.go
```diff
@@ -188,49 +188,81 @@ type Row struct {
...
+	DropMissingTimestamp:    " -- the historian needs a {value, timestamp_ms} payload; the Node-RED JavaScript processor does not add timestamp_ms, so set msg.payload.timestamp_ms = Date.now() (or use the tag processor, which adds it automatically)",
```
led0nk: this also references only those 2 processors, what happens with other processors? we should generalize this message to set the `timestamp_ms` but not with too much detail.

What if you could set thte timestamp with the one from the opcua_server? Then this instruction is not correct as well

historian_plugin/transform.go
```diff
@@ -188,49 +188,81 @@ type Row struct {
...
+	DropContractBypassed:    " -- this versioned data contract carries data_contract_bypassed=true, so its schema was never applied (the registry was unreachable, or no schema is registered for this version) and the payload is unchecked; register the schema or restore the registry, then redeploy. This cannot be overridden by allow_unvalidated_data",
```
led0nk: If you have the historian-plugin running and you register the schema, do we really need to redeploy this plugin here? This should be working as well if this is already running IMO

historian_plugin/transform.go
led0nk: uh and also for all those error messages, let's remove the dashes

historian_plugin/output_historian.go
```diff
@@ -448,22 +443,12 @@ func (o *historianOutput) WriteBatch(ctx context.Context, batch service.MessageB
 	// attribute de-dup cache and make the attribute table grow per-message, so the operator likely
 	// wants them out of metadata_keys.
 	o.warnHighChurnMetadata(churn)
+	if mismatch := drops[DropContractMismatch]; mismatch.count > 0 {
+		o.noteContractMismatch(time.Now(), len(batch), mismatch.count, mismatch.contracts, mismatch.example, len(rows) > 0)
+		return errors.New(nackMessage(o.contract, len(batch), mismatch.count))
```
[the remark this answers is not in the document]
Aaron99B: True. sawMatching is now derived from the drop reason

docs/output/historian.md
```diff
@@ -213,11 +240,12 @@ not representable by the millisecond UNS timestamp and unsuitable for this contr
...
-> change is a realistic operational event rather than a defect. How the plugin should treat a
-> type change there — reject as poison (today), tolerate both `value_num` and `value_text`, or
-> promote the tag to text — is an open policy decision tracked separately; today it is dropped
-> as poison like any other flip.
+> **Generic contracts (`_historian`/`_raw`).** These don't pin a type, so a type change happens in
+> the field and is not a defect. With `allow_unvalidated_data: true` the tag resolves on its natural
+> key alone and the change is accepted: one `topic_id` then holds numeric rows in `value_num` and
+> text rows in `value_text`, with `umh.tag.value_type` recording only the first type seen. Read such
+> a tag with `coalesce(value_num::text, value_text)`. Without the flag a flip is dropped as poison,
+> which is correct for a versioned contract whose schema pins the type.
```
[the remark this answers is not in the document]
Aaron99B: changed that

.changelog/unreleased/breaking-20260730-185013.yaml
```diff
@@ -0,0 +1,3 @@
+kind: breaking
+body: A historian bridge now refuses any batch containing a message for a data contract other than its configured data_contract_name, instead of discarding those messages silently. Nothing in a refused batch is written, so reported write throughput is the number of rows stored rather than the number of messages accepted, and the bridge reports as degraded with an error naming the contracts that arrived and the umh_topics pattern to narrow to. Refused messages are not retried. Bridges subscribed to more than their own contract will store nothing after upgrading until their subscription is narrowed (ENG-5413)
```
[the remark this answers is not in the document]
Aaron99B: changed that

JeremyTheocharis (changes requested):
Can you adfjust the PR description with problem, what we are shipping, what we are not shipping?

.changelog/unreleased/breaking-20260730-185012.yaml
```diff
@@ -0,0 +1,3 @@
+kind: breaking
+body: The historian output reports dropped messages at error level from the first batch onwards, one line per drop reason per batch, naming the reason, how many of the batch it accounted for, an example topic and the fix, so a bridge discarding data no longer looks healthy. Reports are no longer held back for the first 30 seconds, so a bridge fed malformed data reports it immediately instead of appearing healthy while it starts. The reasonless "dropped all N message(s)" warning is gone and the missing_value_or_timestamp reason is split into missing_value and missing_timestamp (ENG-5414)
```
led0nk: there is no kind `breaking` and also i'd prefer to make this like a maximum of 2-3 lines and more precise. also this shouldn't include any direct references to umh-core.

counts for all changelog-entries here

docs/output/historian.md
```diff
@@ -138,10 +139,36 @@ ORDER  BY v.ts DESC;
...
+  `data_contract_bypassed=true`, which the `uns` output sets when the registry was unreachable or no
+  schema is registered, is dropped as `contract_bypassed` with no way to override it. An
+  **unversioned** contract (`_historian`) can never be validated, so it is dropped as
+  `contract_unvalidated` unless you set `allow_unvalidated_data: true` and accept unchecked
+  datatypes.
+- **The payload must be exactly `{value, timestamp_ms}`.** A missing field is dropped as
+  `missing_value` or `missing_timestamp`, an extra top-level field as `not_timeseries`, which is what
+  refuses a relational record. Neither rule is configurable. The
+  [tag processor](../processing/tag-processor.md) fills `timestamp_ms`; the
+  [Node-RED JavaScript processor](../processing/node-red-javascript-processor.md) does not, so a
```
led0nk: as said we shouldn't mention, which processors doesn't set this. Personally i'd mention that this is automatically set via the tag-processor otherwise you need to ensure that the payload holds a `timestamp_ms` as well.

docs/output/historian.md
```diff
@@ -138,10 +139,36 @@ ORDER  BY v.ts DESC;
...
+  write flow that reshapes the payload has to carry it across.
+- **A malformed message is dropped, never nacked.** An unparseable `umh_topic`
+  (`invalid_topic`), a non-finite number (`unclassifiable_value`), an unreadable timestamp
+  (`bad_timestamp`) and a payload that is not a JSON object (`not_structured`, `not_object`) are
+  skipped; the rest of the batch is written. Each reason logs one error per batch with its share of
+  the batch, an example topic and its fix, which umh-core reads as a degraded bridge. The log clears
+  once the bad messages stop and the last error ages out of umh-core's window.
```
led0nk: Also don't include all those umh-core references here, this should be seen as a "standalone" (not our thingy) plugin which is usable with plain benthos

docs/output/historian.md
```diff
@@ -244,7 +273,9 @@ On top of benthos's built-in output metrics (`output_sent`, `output_error`,
 - `historian_value_rows_written` — value rows upserted (counted after the batch commits).
 - `historian_attribute_rows_written` — attribute rows upserted; the gap below the value-row
   count is metadata de-duplication at work.
-- `historian_messages_dropped` (labelled by `reason`) — messages dropped before any write.
+- `messages_dropped` (labelled by `reason`) — messages dropped before any write. The name has no
+  plugin prefix because every benthos-umh plugin uses the same drop counter, which is how umh-core
```
led0nk: no umh-core mentions here as well, could be some generic information here, like you could correct metrics like this...

docs/processing/node-red-javascript-processor.md
```diff
@@ -6,6 +6,16 @@ Use the `nodered_js` processor instead of the `tag_processor` when you need full
 
 For the full list of available JavaScript globals (`msg`, `console`, `cache`, `protobuf`), see the [JavaScript API Reference](javascript-api.md).
 
+> **This processor does not add `timestamp_ms`.** It emits exactly the payload your code returns.
```
led0nk: As discussed, let's remove this, we don't want to mention everywhere what something doesn't do

docs/output/historian.md
```diff
@@ -28,6 +28,7 @@ No JavaScript processor or hand-written `sql_raw` is needed.
 | `password` | yes | — | Role password (plaintext in config; redacted in logs). |
 | `sslmode` | no | `require` | `require` \| `disable` \| `verify-full`. |
 | `sslrootcert` / `sslcert` / `sslkey` | no | `""` | TLS cert paths inside the container. |
+| `allow_unvalidated_data` | no | `false` | Store data from an **unversioned** contract (e.g. `_historian`), accepting that its datatypes are never schema-checked. No effect on a versioned contract, where a bypassed schema is always rejected. |
```
led0nk: let's talk about this again, then you could resolve based on discussion

historian_plugin/mismatch.go
```diff
@@ -0,0 +1,65 @@
...
+		out = append(out, c)
+	}
+	sort.Strings(out)
+	return out
+}
+
+func mismatchMessage(contract string, overBroad bool, total int, mismatched int, contracts []string, example string) string {
+	seen := "[" + strings.Join(contracts, ", ") + "]"
+	if overBroad {
+		return fmt.Sprintf("TimescaleDB historian: subscription is over-broad (reason=%s) -- %d of %d message(s) in this batch carry other data contracts %s (example: %s), so the whole batch is refused and any matching message(s) are not written either. Narrow umh_topics to %s",
```
led0nk: lets narrow this down, i'd prefer to not use the dashes "--"
Also I'd prefer to make the errormessages here as short & precise as possible.

What's important here? imo:
- dropreason
- how many of x messages
- can also be suggestedTopicPattern

i'd let go of examples, as well as seen data-contracts here

historian_plugin/mismatch.go
```diff
@@ -0,0 +1,65 @@
...
+			DropContractMismatch, mismatched, total, seen, example, suggestedTopicPattern(contract))
+	}
+	return fmt.Sprintf("TimescaleDB historian: no message carries data contract _%s (reason=%s); the subscription selects %s (example: %s) and every batch is refused. Either set data_contract_name to a contract that is published, or narrow umh_topics to %s",
```
led0nk: same for this here

historian_plugin/mismatch.go
```diff
@@ -0,0 +1,65 @@
...
+		contract, DropContractMismatch, seen, example, suggestedTopicPattern(contract))
+}
+
+func nackMessage(contract string, total int, mismatched int) string {
+	return fmt.Sprintf("TimescaleDB historian: batch refused, %d of %d message(s) do not carry data contract _%s (reason=%s)",
+		mismatched, total, contract, DropContractMismatch)
+}
+
+func (o *historianOutput) noteContractMismatch(now time.Time, total int, mismatched int, seen map[string]struct{}, example string, sawMatching bool) {
+	o.logStateMu.Lock()
```
led0nk: i'll need an explanation on this and why we need this, maybe i'm mislead here.
But thinking of this, maybe we need to add then some code-explanation

docs/output/historian.md
led0nk: also an additional thought on this, (not being addressed in here) when we state that historian is never schema-validated, maybe we should then add this to redpanda as well by default? (just a random thought)

historian_plugin/output_historian.go
```diff
@@ -402,33 +402,35 @@ func (o *historianOutput) WriteBatch(ctx context.Context, batch service.MessageB
...
 		// Transform validates the umh_topic/contract and the value+timestamp, and decides whether
 		// this row also needs to write a metadata (attribute) row. A non-empty reason means drop.
-		row, reason := Transform(payload, meta, o.contract, o.metadataKeysAll, o.metadataKeys, o.metadataExclude, view)
-		if reason != DropNone {
-			if reason == DropContractMismatch {
-				contractMismatches++
-			}
-			o.recordDrop(string(reason), meta["umh_topic"])
+		row, drop := Transform(payload, meta, o.contract, o.metadataKeysAll, o.metadataKeys, o.metadataExclude, o.allowUnvalidated, view)
+		// Transform checks topic, then contract, then payload: any later reason means the contract
```
led0nk: can we further state why we need this, than what it does?

historian_plugin/output_historian.go
```diff
@@ -503,14 +488,23 @@ func (o *historianOutput) WriteBatch(ctx context.Context, batch service.MessageB
...
-	err := pool.QueryRow(ctx, topicLookupSQL, r.RawLocation, r.ContractName, r.VirtualPath, r.TagName, string(r.ValueType)).Scan(&id)
+	var err error
+	if o.allowUnvalidated {
+		err = pool.QueryRow(ctx, topicLookupAnyTypeSQL, r.RawLocation, r.ContractName, r.VirtualPath, r.TagName).Scan(&id)
+	} else {
+		err = pool.QueryRow(ctx, topicLookupSQL, r.RawLocation, r.ContractName, r.VirtualPath, r.TagName, string(r.ValueType)).Scan(&id)
+	}
+	resolve := topicResolveSQL
+	if o.allowUnvalidated {
+		resolve = topicResolveKeepTypeSQL
```
led0nk: does it make sense to put this in the branch above where we already check on the flag?

led0nk (changes requested):
lets maybe go through this, if it fits for you

.changelog/unreleased/breaking-20260730-185012.yaml
Aaron99B: Changed that

docs/output/historian.md
Aaron99B: Changed that

docs/output/historian.md
Aaron99B: changed that

docs/processing/node-red-javascript-processor.md
Aaron99B: removed it

historian_plugin/mismatch.go
Aaron99B: did this

historian_plugin/transform.go
Aaron99B: adjusted the error message

docs/output/historian.md
Aaron99B: did that

historian_plugin/mismatch.go
Aaron99B: changed it

historian_plugin/mismatch.go
Aaron99B: The idea was to not note it all the time in the logs, but to have it throttled

historian_plugin/output_historian.go
Aaron99B: it is for checking if the configured contract is published

historian_plugin/output_historian.go
Aaron99B: yes, did that

historian_plugin/transform.go
Aaron99B: changed that

.changelog/unreleased/improvements-20260730-185015.yaml
```diff
@@ -0,0 +1,3 @@
+kind: improvements
+body: The historian output reports dropped messages at error level, one aggregated line per reason per batch, naming the reason, how much of the batch it accounted for, an example topic and the fix. Reporting starts with the first batch rather than after 30 seconds, and missing_value_or_timestamp is split into missing_value and missing_timestamp (ENG-5414)
```
Aaron99B: Janik: add code quotas for parameter names

docs/output/historian.md
Aaron99B: Changed it such that we do not do a detection of contracts like _historian and _raw anymore. The behaviour now is: users can store values in these contracts. However if there are datatype changes, the data will get dropped and an error gets raised. If now users set the new flag allow_dataype_changes to true, any data where the datatype flips can be saved

.changelog/unreleased/improvements-20260730-185015.yaml
Aaron99B: did that

historian_plugin/transform.go
Aaron99B: adjusted the wording

historian_plugin/mismatch.go
Aaron99B: We just need it for the to check when we logged the last mitmatch due to the parallelisation we have. I adjusted the naming and removed the mutex from the check if we ever logged anything, because for a one word bool we do not need a mutex

led0nk (approved):

JeremyTheocharis (approved):

PR #436, opened 2026-08-10
Modbus: read an address from a subset of slaves

RuneRoven:
## Problem

With `slaveIDs: [1, 2, 3, 4, 5, 6]`, an address could be read from exactly one slave or from all six — never from a subset. Anyone wanting register 23 from slaves 3, 5 and 6 had to duplicate the address entry three times under three different names, which is also how those tags then showed up in the UNS.

## What we are shipping

- The existing `slaveID` option in a `unifiedAddresses` string now takes a comma-separated list: `cycle_count.holding.23.INT16:slaveID=3,5,6`. No new config key, so the Management Console form is structurally unchanged — the config-spec JSON diff against staging is one description sentence plus one example entry.
- `0` alone still means all slaves. `0` inside a list, a repeated ID, an empty element, and any ID missing from the top-level `slaveIDs` are all rejected at startup rather than silently reading nothing.
- Every listed slave emits its own message under the same `modbus_tag_name`, differing only in `modbus_tag_slaveid`. A pipeline that builds its topic from the tag name alone would collapse them onto one topic, so the plugin logs a startup warning and the docs show a `tag_processor` snippet that puts the slave ID in the `virtual_path`.
- Two test-quality fixes found while doing this: `modbus_test.go` had its own copy of the distribution and per-slave dedup logic that would have silently drifted from production, and the second-pass dedup was unreachable from any test. Both now call the real code.

## What we are NOT shipping

- The deprecated `addresses` object list is untouched — it still takes a single `slaveID` int. Subsets are `unifiedAddresses`-only.
- No `{slaveID}` placeholder in the name segment. It would remove the need for the `tag_processor` snippet above, but it grows the address grammar and is a separate decision.
- `ModbusDataItemWithAddress.SlaveID byte` is kept as-is with `SlaveIDs []byte` added alongside, rather than collapsing both into one field. The collapse is cleaner but breaks the exported struct.
- `docs/input/modbus.md` currently says the startup warning fires whenever an address is read from more than one slave; it actually fires only for an explicit multi-ID list, not the all-slaves case. Wording fix, happy to take it in review.

Fixes ENG-4688

### DIFF
--- a/.changelog/unreleased/new-20260806-171145.yaml
+++ b/.changelog/unreleased/new-20260806-171145.yaml
@@ -1,0 +1,3 @@
+kind: new
+body: Modbus addresses can be restricted to a subset of slaves with slaveID=3,5,6 in unifiedAddresses (ENG-4688)
+time: 2026-08-06T17:11:45.873311+02:00
--- a/docs/input/modbus.md
+++ b/docs/input/modbus.md
@@ -266,7 +266,8 @@ name.register.address.type[:key=value]*
 4. `type` — one of: `BIT`, `INT8L`, `INT8H`, `UINT8L`, `UINT8H`, `INT16`, `UINT16`, `INT32`, `UINT32`, `INT64`, `UINT64`, `FLOAT16`, `FLOAT32`, `FLOAT64`, `STRING`
 
 **Optional key-value pairs** (colon-separated):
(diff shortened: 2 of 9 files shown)


JeremyTheocharis:
tagging @rpaland for any UI implications here

JeremyTheocharis:
I have closed it for now as there is no capacity at the moment to review it. Let's repitch it internally first.

PR #437, opened 2026-08-17
feat(historian): keep the data contract version in the stored metadata

Aaron99B:
## Problem

`data_contract_version` never reached the database. It was in the historian's structural
skip list, which is for keys already stored as a column. The version has no column: all
versions of a contract share one `umh.tag` row.

## What we are shipping

- The version is stored as metadata. Query it with `attribute->>'data_contract_version'`.
- A contract with no version, like `_historian`, stores no version key.
- Moving a tag from `_pump_v1` to `_pump_v2` writes a new attribute row.

## What we are NOT shipping

No column, no migration. The metadata allowlist could do this too, but only if you turn off
`metadata_keys_all` and list every key by hand, so the fix goes in the plugin.

Fixes ENG-5474


JeremyTheocharis (commented):
would be nice to have a reduced PR description :)

JeremyTheocharis (approved):

led0nk (approved):

PR #441, opened 2026-08-20
Configurable chunk interval for the historian hypertables

Aaron99B:
## Problem

Both historian hypertables were created with a hardcoded 7-day chunk width, which sets how much a
time-filtered query can skip and how coarsely compression and retention act. Changing it needed psql
and `set_chunk_time_interval`, documented nowhere in this repo.

## What we are shipping

- `value_chunk_interval` and `attribute_chunk_interval`, defaulting to `168h` — the width the tables
  already had, so a fresh bootstrap is unchanged.
- Applied once at table creation, like `compress_after` and `retention`, with a restart warning when
  the config and the table no longer agree.
- A docs section on changing the interval on a live database.

## What we are NOT shipping

No Management Console wizard field, and no migration for existing tables — an interval change still
only affects chunks created after it.

Fixes ENG-5689


JeremyTheocharis:
added yannic to learn from janik on how to review it

docs/output/historian.md
```diff
@@ -358,6 +360,27 @@ After changing a policy on the database, set the **same** value in the bridge co
...
+query can skip, and how coarsely compression and retention act, since both work on whole chunks.
+
+The interval reaches the database only when the table is created, and for the same reason the
+policies do: a config edit should not silently reorganize production history. On restart the bridge
+warns when a configured interval differs from the one its table was created with.
+
+To change it on an existing database, for a contract named `pump`:
+
+```sql
+SELECT set_chunk_time_interval('umh.value_pump', INTERVAL '1 day');
```
led0nk, line 377: think this should represent the current default-values, which is for values 7 days as well

historian_plugin/output_historian.go
```diff
@@ -52,6 +52,8 @@ func historianConfig() *service.ConfigSpec {
...
 		Field(service.NewStringField("compress_after").Description("Compress chunks older than this, as a Go duration; use hours (e.g. \"168h\") -- days are not a valid unit. Applied once at first database bootstrap. Per contract.").Default("168h").Advanced()).
 		Field(service.NewStringField("retention").Description("Drop chunks older than this, as a Go duration; use hours (e.g. \"720h\") -- days are not a valid unit. Empty = keep forever. Applied once at first database bootstrap.").Default("").Advanced()).
+		Field(service.NewStringField("value_chunk_interval").Description("Chunk width of the value hypertable, as a Go duration; use hours (e.g. \"168h\") -- days are not a valid unit. Applied once when the table is created; changing it later needs set_chunk_time_interval on the database and affects only new chunks.").Default("168h").Advanced()).
```
led0nk: description should be refactored imo, because it says sth like "Go duration" (what is meant by that?)  as well as those "--". Should be short but descriptive.

historian_plugin/output_historian.go
```diff
@@ -52,6 +52,8 @@ func historianConfig() *service.ConfigSpec {
...
+		Field(service.NewStringField("attribute_chunk_interval").Description("Chunk width of the attribute hypertable, as a Go duration. Attribute rows are written only when a tag's metadata changes, so this table is far sparser than the value table. Applied once when the table is created.").Default("168h").Advanced()).
```
led0nk: same here as well

historian_plugin/output_historian.go
```diff
@@ -385,6 +410,26 @@ func (o *historianOutput) warnPolicyDrift(ctx context.Context) {
...
+	return applied
+}
+
+// warnChunkDrift warns when a table's chunk interval differs from config. create_hypertable is a
+// no-op once the table exists, so editing an interval and restarting otherwise has no visible effect.
+func (o *historianOutput) warnChunkDrift(ctx context.Context) {
+	appliedValue := o.readChunkInterval(ctx, "value_"+o.contract)
+	appliedAttribute := o.readChunkInterval(ctx, "attribute_"+o.contract)
+	for _, w := range chunkDriftWarnings(int64(o.valueChunk.Seconds()), appliedValue, int64(o.attributeChunk.Seconds()), appliedAttribute) {
+		o.logger.Warnf("TimescaleDB historian: %s. The chunk interval is set when the table is created and never re-applied, so a config change alone does not update it. Change it on the database with set_chunk_time_interval on umh.value_%s and umh.attribute_%s -- existing chunks keep the width they were created with -- then set the same value in this output's config to silence this warning.", w, o.contract, o.contract)
```
led0nk: let's focus on clear & precise log messages, we don't want to overwhelm the user

historian_plugin/sql.go
```diff
@@ -367,12 +367,53 @@ func policyDriftWarnings(compressWant int64, appliedComp *int64, retentionWant *
...
+	retentionSet   bool
+	valueChunk     time.Duration
+	attributeChunk time.Duration
+}
+
+// The chunk interval needs no ledger gate: create_hypertable with if_not_exists => TRUE is a no-op
+// once the table is a hypertable, so a changed interval never retro-applies. Consequence: an edit
+// reaches an existing database only via set_chunk_time_interval, and only for new chunks.
+// chunkIntervalSQL reads the chunk width TimescaleDB has applied to a umh hypertable, in seconds,
+// as a scalar subquery so it always yields one row (NULL when the table is not a hypertable here).
```
led0nk: please reduce comment size, only if needed state `why` we need this

historian_plugin/output_historian.go
```diff
@@ -385,6 +410,26 @@ func (o *historianOutput) warnPolicyDrift(ctx context.Context) {
 	}
 }
 
+// readChunkInterval reads the chunk width applied to one hypertable, in seconds. Best-effort: an
+// unreadable catalog returns nil, which warnChunkDrift treats as "say nothing".
+func (o *historianOutput) readChunkInterval(ctx context.Context, table string) *int64 {
+	var applied *int64
+	if err := o.pool.QueryRow(ctx, chunkIntervalSQL(), table).Scan(&applied); err != nil {
+		return nil
```
led0nk, line 432: pretty good example where we would actually need a comment.
question that comes up is: why do we suppress the error here? what if the query fails?
+ 
since this is a pointer, let's doublecheck what happens if we actually return nil here

historian_plugin/output_historian.go
```diff
@@ -246,8 +246,32 @@ func (o *historianOutput) dsn() string {
...
+	if err != nil {
+		return 0, fmt.Errorf("%s: %w", field, err)
+	}
+	if parsed < time.Second {
+		return 0, fmt.Errorf("%s must be at least 1s, got %q", field, raw)
+	}
+	return parsed, nil
+}
+
 func (o *historianOutput) bootstrapStmt() string {
```
led0nk: can we make this a valuable function-name?

historian_plugin/sql.go
```diff
@@ -367,12 +367,53 @@ func policyDriftWarnings(compressWant int64, appliedComp *int64, retentionWant *
...
+	if appliedValue != nil && *appliedValue != valueWant {
+		warns = append(warns, fmt.Sprintf("configured value_chunk_interval (%ds) does not match the chunk interval the value hypertable was created with (%ds)", valueWant, *appliedValue))
+	}
+	if appliedAttribute != nil && *appliedAttribute != attributeWant {
+		warns = append(warns, fmt.Sprintf("configured attribute_chunk_interval (%ds) does not match the chunk interval the attribute hypertable was created with (%ds)", attributeWant, *appliedAttribute))
+	}
+	return warns
+}
+
+func bootstrapSQL(cfg bootstrapConfig) string {
```
led0nk, line 407: maybe instead of replacing strings here we could make use of a map or sth like this

docs/output/historian.md
```diff
@@ -35,6 +35,8 @@ No JavaScript processor or hand-written `sql_raw` is needed.
 | `metadata_keys_exclude` | no | `[]` | Blacklist applied only when `metadata_keys_all=true`. Each entry is an exact key name or a trailing-`*` prefix (e.g. `opcua_*`); matches are dropped on top of the built-in exclusions. A bare `*` drops everything. Ignored in allowlist mode. |
 | `compress_after` | no | `168h` | Compress chunks older than this. **Applied once at first bootstrap** — changing it in the bridge afterward has no effect (see [Changing compression or retention](#changing-compression-or-retention)). |
 | `retention` | no | `""` | Drop chunks older than this; empty keeps data forever. **Applied once at first bootstrap** — changing it in the bridge afterward has no effect (see [Changing compression or retention](#changing-compression-or-retention)). |
+| `value_chunk_interval` | no | `168h` | Chunk width of `umh.value_<contract>`. **Applied once, when the table is created** — changing it in the bridge afterward has no effect (see [Changing the chunk interval](#changing-the-chunk-interval)). |
```
led0nk: again i didn't see this last time, but this is also referring to bridges, we don't do that within benthos-umh here, applies for all those comments basically

historian_plugin/output_historian.go
```diff
@@ -246,8 +246,32 @@ func (o *historianOutput) dsn() string {
 	return o.buildDSN()
 }
 
+// Sub-second durations render as INTERVAL '0 seconds' (whole-second SQL) and make an invalid
+// policy or chunk width; reject here for a clear error instead of a bootstrap failure.
+func wholeSecondDuration(conf *service.ParsedConfig, field string) (time.Duration, error) {
```
led0nk: lets rename this to sth like parse... which fits more imo

led0nk (changes requested):

docs/output/historian.md
Aaron99B: changed that

docs/output/historian.md
Aaron99B: adapted it to state the defaults, but the values here differ form the defaults on purpose to show how one can change them

historian_plugin/output_historian.go
Aaron99B: changed that

historian_plugin/output_historian.go
Aaron99B: changed

historian_plugin/output_historian.go
Aaron99B: did that

historian_plugin/output_historian.go
```diff
@@ -246,8 +246,32 @@ func (o *historianOutput) dsn() string {
...
+	}
+	parsed, err := time.ParseDuration(raw)
+	if err != nil {
+		return 0, fmt.Errorf("%s: %w", field, err)
+	}
+	if parsed < time.Second {
+		return 0, fmt.Errorf("%s must be at least 1s, got %q", field, raw)
+	}
+	return parsed, nil
+}
```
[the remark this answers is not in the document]
Aaron99B: changed that

historian_plugin/output_historian.go
Aaron99B: did that

historian_plugin/output_historian.go
Aaron99B: included it

historian_plugin/output_historian.go
Aaron99B: changed that

historian_plugin/sql.go
Aaron99B: did that

historian_plugin/sql.go
Aaron99B: nice idea. Using a map now

docs/output/historian.md
```diff
@@ -33,8 +33,10 @@ No JavaScript processor or hand-written `sql_raw` is needed.
...
+| `compress_after` | no | `168h` | Compress chunks older than this. **Applied once at first bootstrap** — changing it afterward has no effect (see [Changing compression or retention](#changing-compression-or-retention)). |
+| `retention` | no | `""` | Drop chunks older than this; empty keeps data forever. **Applied once at first bootstrap** — changing it afterward has no effect (see [Changing compression or retention](#changing-compression-or-retention)). |
+| `value_chunk_interval` | no | `168h` | Chunk width of `umh.value_<contract>`. **Applied once, when the table is created** — changing it afterward has no effect (see [Changing the chunk interval](#changing-the-chunk-interval)). |
+| `attribute_chunk_interval` | no | `168h` | Chunk width of `umh.attribute_<contract>`. Attribute rows are written only when a tag's metadata changes, so this table is far sparser than the value table and rarely needs a different width. **Applied once, when the table is created.** |
```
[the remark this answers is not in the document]
Aaron99B: we already have the default values as examples

historian_plugin/output_integration_test.go
```diff
@@ -94,6 +94,35 @@ var _ = Describe("TimescaleDB integration", Ordered, Label("postgres"), func() {
...
+		restarted.SetChunkIntervals(24*time.Hour, 168*time.Hour)
+		logs := restarted.CaptureLogs()
+		Expect(restarted.Connect(ctx)).To(Succeed())
+		defer restarted.Close(ctx)
+
+		Expect(logs()).To(ContainSubstring("configured value_chunk_interval (86400s)"))
+		Expect(logs()).To(ContainSubstring("created with (604800s)"))
+		Expect(logs()).To(ContainSubstring("stays in force"))
+		Expect(logs()).NotTo(ContainSubstring("attribute_chunk_interval"))
+		Expect(restarted.AppliedChunkInterval(ctx, "value_chunkdrift")).To(HaveValue(Equal(int64(604800))))
```
FalconTube: Same as above

historian_plugin/export_test.go
```diff
@@ -370,3 +377,32 @@ func (h *HistorianTestHandle) ReportDropsForTest(total int, summaries map[string
...
+// oracle.
+func (h *HistorianTestHandle) AppliedChunkInterval(ctx context.Context, table string) *int64 {
+	ExpectWithOffset(1, h.o.pool).NotTo(BeNil(), "Connect must succeed before AppliedChunkInterval")
+	var applied *int64
+	err := h.o.pool.QueryRow(ctx,
+		"SELECT EXTRACT(EPOCH FROM time_interval)::bigint FROM timescaledb_information.dimensions WHERE hypertable_schema = 'umh' AND hypertable_name = $1 AND column_name = 'ts'",
+		table).Scan(&applied)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	return applied
+}
```
FalconTube, line 401: Shouldn't this use the real, existing `readChunkInterval` function?

historian_plugin/output_historian.go
```diff
@@ -381,7 +407,37 @@ func (o *historianOutput) warnPolicyDrift(ctx context.Context) {
...
+	}
+	appliedAttribute, err := o.readChunkInterval(ctx, "attribute_"+o.contract)
+	if err != nil {
+		o.logger.Warnf("historian: cannot read the chunk interval of umh.attribute_%s: %v", o.contract, err)
+		return
+	}
+	for _, w := range chunkDriftWarnings(int64(o.valueChunk.Seconds()), appliedValue, int64(o.attributeChunk.Seconds()), appliedAttribute) {
+		o.logger.Warnf("historian: %s. The applied width stays in force, and existing chunks keep theirs.", w)
 	}
 }
```
FalconTube, line 445: Looking at the code coverage, only the happy paths are currently tested.<br>Pls add tests for the error cases as well.

historian_plugin/output_historian.go
```diff
@@ -315,10 +340,11 @@ func (o *historianOutput) Connect(ctx context.Context) error {
 			return err
 		}
 		defer conn.Release()
-		if _, err := conn.Exec(bootCtx, o.bootstrapStmt()); err != nil {
+		if _, err := conn.Exec(bootCtx, o.renderBootstrapDDL()); err != nil {
 			return fmt.Errorf("schema bootstrap failed: %w", err) // guard stays false -> next Connect retries
 		}
 		o.warnPolicyDrift(bootCtx)
```
FalconTube, line 346: The `o.warnPolicyDrift(bootCtx)` function is basically not covered by tests.<br>Wasn't this exactly the issue about non-applied policies that was recently reported?<br><br>Might make sense to add tests for it

FalconTube (changes requested):
Already much better after review from led0nk.<br>Looks quite good now, just some smaller notes

historian_plugin/output_integration_test.go
```diff
@@ -94,6 +94,36 @@ var _ = Describe("TimescaleDB integration", Ordered, Label("postgres"), func() {
...
 
+	It("creates each hypertable with its own configured chunk interval", func() {
+		h := tsh.NewHistorianTestHandle(sharedDSN, "chunky")
+		h.SetChunkIntervals(24*time.Hour, 720*time.Hour)
+		Expect(h.Connect(ctx)).To(Succeed())
+		defer h.Close(ctx)
+
+		Expect(h.AppliedChunkInterval(ctx, "value_chunky")).To(HaveValue(Equal(int64(86400))))
+		Expect(h.AppliedChunkInterval(ctx, "attribute_chunky")).To(HaveValue(Equal(int64(2592000))))
+	})
```
FalconTube, line 110: In here, it is not obvious what the time is.<br>Better declare the two times as variables and then pass them into the setup and the `Expect` functions.<br>This makes it very clear, that the `Expect` step just tests the correctness of the setup function

historian_plugin/export_test.go
Aaron99B: True. Changed that

historian_plugin/output_historian.go
Aaron99B: good point. added tests for it

historian_plugin/output_historian.go
Aaron99B: Did that now

historian_plugin/output_integration_test.go
Aaron99B: Nice idea. Moved it to constants

historian_plugin/output_integration_test.go
Aaron99B: Moved it to constants as well

FalconTube (approved):
Good to go from my side now!

led0nk (approved):

PR #442, opened 2026-08-26
fix(historian): apply compression and retention to every data contract

Aaron99B:
Stacked on #441.

## Problem

Compression and retention only ever reached the first historian data contract on a database. The
policy block was gated on a row in `umh.schema_migrations`, which exists once per database, while
the hypertables it configures exist once per contract. Every later contract got its tables and no
policies, so its chunks were never compressed and old ones never dropped, with nothing in the logs
saying so.

## What we are shipping

- Every contract gets its own policies, including one added to a database that already holds others.
  A contract missing them gets them the next time its bridge starts.
- Policies are only ever added, never removed or altered. Retention is added only to a hypertable
  with no rows in it: on one that already holds rows the output logs the `add_retention_policy`
  statement to run instead, so no restart can delete stored data.
- A policy someone removes with `remove_retention_policy` is therefore not re-created while that
  hypertable holds rows, whatever `retention` says. On an empty hypertable the next start does
  re-create it, so clear `retention` in the config there.
- An interval an operator set on the database is never overwritten. A divergence is reported per
  hypertable, on both of a contract's tables.
- Docs and field descriptions match the behaviour, including the compatibility note: the
  compression-settings `ALTER` does run against an already-created table, taking a brief exclusive
  lock on it.

## What we are NOT shipping

- Retention on a hypertable that already holds rows. That stays an `add_retention_policy` someone
  runs deliberately; the output only prints it.
- A way to switch compression off. `compress_after` has no off value, so a compression policy
  removed on the database returns at the next start. An off value would also break the drift check's
  catalog probe, which relies on compression always having a policy, so it needs its own design.
- Changing an interval from the config: still no effect on a hypertable that already has that
  policy, and the drift warning is still the only signal.
- A physical name mapping for long contracts. A name over 53 characters is now refused at config
  parse, because PostgreSQL truncates `attribute_<contract>` at 63 bytes and two contracts differing
  only in the truncated tail would share one pair of tables. Supporting longer names is ENG-5765.

Fixes ENG-5757

### DIFF
--- a/.changelog/unreleased/fixes-20260826-154709.yaml
+++ b/.changelog/unreleased/fixes-20260826-154709.yaml
@@ -1,3 +1,3 @@
 kind: fixes
-body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. Retention reaches only a hypertable that holds no data yet, so upgrading never deletes stored rows, and for a contract that already holds data the output logs the add_retention_policy statement to run by hand (ENG-5757)
+body: The historian output now applies compression and retention to every data contract. Previously it was only applied to the first one. This is backwards-compatible with existing historian-setups. (ENG-5757)
 time: 2026-08-26T15:47:09.959042+02:00
--- a/.changelog/unreleased/fixes-20260826-154709.yaml
+++ b/.changelog/unreleased/fixes-20260826-154709.yaml
@@ -1,3 +1,3 @@
 kind: fixes
-body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. A contract still missing its policies gets them the next time the output starts, and if retention is set, TimescaleDB then drops chunks older than that window, so check retention against the policies already applied before upgrading (ENG-5757)
+body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. Retention is applied only to a hypertable that holds no data yet, so upgrading never deletes history: for a contract that already holds data the output logs the add_retention_policy statement to run by hand (ENG-5757)
 time: 2026-08-26T15:47:09.959042+02:00
--- a/.changelog/unreleased/fixes-20260826-154709.yaml
+++ b/.changelog/unreleased/fixes-20260826-154709.yaml
@@ -1,3 +1,3 @@
 kind: fixes
-body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. A contract still missing its policies gets them the next time its bridge starts, and if retention is set, TimescaleDB then drops chunks older than that window, so check retention against the policies already applied before upgrading (ENG-5757)
+body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. A contract still missing its policies gets them the next time the output starts, and if retention is set, TimescaleDB then drops chunks older than that window, so check retention against the policies already applied before upgrading (ENG-5757)
 time: 2026-08-26T15:47:09.959042+02:00
--- a/.changelog/unreleased/fixes-20260826-154709.yaml
+++ b/.changelog/unreleased/fixes-20260826-154709.yaml
@@ -1,0 +1,3 @@
+kind: fixes
+body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. A contract still missing its policies gets them the next time its bridge starts, and if retention is set, TimescaleDB then drops chunks older than that window, so check retention against the policies already applied before upgrading (ENG-5757)
+time: 2026-08-26T15:47:09.959042+02:00
--- a/.changelog/unreleased/fixes-20260826-154709.yaml
+++ b/.changelog/unreleased/fixes-20260826-154709.yaml
@@ -1,3 +1,3 @@
 kind: fixes
-body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. Retention is applied only to a hypertable that holds no data yet, so upgrading never deletes history: for a contract that already holds data the output logs the add_retention_policy statement to run by hand (ENG-5757)
+body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. Retention reaches only a hypertable that holds no data yet, so upgrading never deletes stored rows, and for a contract that already holds data the output logs the add_retention_policy statement to run by hand (ENG-5757)
 time: 2026-08-26T15:47:09.959042+02:00
--- a/historian_plugin/output_historian.go
+++ b/historian_plugin/output_historian.go
@@ -347,7 +347,7 @@ func (o *historianOutput) Connect(ctx context.Context) error {
 			return err
 		}
 		defer conn.Release()
-		o.warnRetentionRecreate(bootCtx)
+		o.warnRetentionAboutToDrop(bootCtx)
 		if _, err := conn.Exec(bootCtx, o.renderBootstrapDDL()); err != nil {
 			return fmt.Errorf("schema bootstrap failed: %w", err) // guard stays false -> next Connect retries
 		}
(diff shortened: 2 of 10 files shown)


historian_plugin/output_historian.go
```diff
@@ -305,6 +306,18 @@ func (o *historianOutput) Connect(ctx context.Context) error {
...
+		// without this the decline reaches no log.
+		cfg.ConnConfig.OnNotice = func(_ *pgconn.PgConn, n *pgconn.Notice) {
+			// WARNING is what add_*_policy raises when if_not_exists makes it decline; NOTICE is
+			// routine bootstrap chatter ("trigger ... does not exist, skipping") on every start.
+			if n.Severity == "WARNING" {
+				o.logger.Warnf("historian: %s", n.Message)
+				return
+			}
+			o.logger.Debugf("historian: %s: %s", n.Severity, n.Message)
+		}
```
FalconTube, line 324: I don't quite understand why we need this function?<br>Also, if the severity level is e.g. ERROR, it will be logged as DEBUG. Why?

historian_plugin/output_historian.go
```diff
@@ -376,38 +390,72 @@ func (o *historianOutput) probeWritable(ctx context.Context) error {
...
+			continue
+		}
+		var applied *int64
+		if err := o.pool.QueryRow(ctx, policyIntervalSQL("policy_retention", "drop_after"), table).Scan(&applied); err != nil {
+			o.logger.Warnf("historian: cannot read the retention policy of umh.%s: %v", table, err)
+			continue
+		}
+		if applied != nil {
+			continue
+		}
```
FalconTube: Here the two error paths are not covered by tests?<br>Is there a reason for it?<br>Otherwise lets add a small test for them :)

historian_plugin/output_historian.go
```diff
@@ -376,38 +390,72 @@ func (o *historianOutput) probeWritable(ctx context.Context) error {
...
+// not reported until it next starts. Best-effort: a failed read is logged and never fails Connect.
+func (o *historianOutput) warnPolicyDrift(ctx context.Context, retentionNotApplied map[string]struct{}) {
 	var retentionWant *int64
 	if o.retentionSet {
 		v := int64(o.retention.Seconds())
 		retentionWant = &v
 	}
-	for _, w := range policyDriftWarnings(int64(o.compressAfter.Seconds()), appliedComp, retentionWant, appliedRet) {
-		o.logger.Warnf("historian: %s. The applied value stays in force.", w)
+	for _, table := range []string{"value_" + o.contract, "attribute_" + o.contract} {
```
FalconTube: This declaration is re-used in multiple parts of this file.<br>Makes sense to define it once globally and then re-use it.<br><br>Also, if something about these values ever changes in `sql.go`, then the compiler will not catch this.<br>But I think there is no nice way around it.<br><br>Still, with the single value declaration, it will be easier to maintain

historian_plugin/output_integration_test.go
```diff
@@ -213,6 +210,187 @@ var _ = Describe("TimescaleDB integration", Ordered, Label("postgres"), func() {
 		Expect(h.SchemaVersion(ctx)).To(Equal(1))
 	})
 
+	It("applies compression and retention to a contract added after the database was bootstrapped", func() {
+		first := tsh.NewHistorianTestHandle(sharedDSN, "polfirst")
+		first.SetPolicies(168*time.Hour, 720*time.Hour, true)
+		Expect(first.Connect(ctx)).To(Succeed())
+		defer first.Close(ctx)
```
FalconTube: I think the `Close` statement should not be `defered` here.<br>Because with the `defer`, we have two simultaneous connections.<br><br>But in the real world, "old" benthos has already set up the policies and then "new" benthos version does the requested changes. <br>Then the two connections are not simultaneous but sequential<br><br>This same pattern is also present in other tests here, might make sense to unify it everywhere.<br>Here short overview from Klaus:

```
`:179` — "a fresh handle connects to an already-bootstrapped database... (restart path)"
- `:258`, `:272` — "gives a contract left without policies by an older build..." 
- `:298`, `:309` — "leaves a retention policy removed..."
- `:324`, `:332` — "keeps an operator's own retention interval across a restart"
- `:344`, `:355` — "re-adds a removed compression policy..."
- `:461`, `:466` — "re-warms after restart" (comment literally says "fresh handle == restart path")
```

historian_plugin/output_integration_test.go
FalconTube, line 1: Also please set declaration and expected time values as constants in the tests, as we did in the previous PR 🙂

FalconTube (changes requested):
Just some smaller code issues, then good for a quick re-review

.changelog/unreleased/fixes-20260826-154709.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. Retention is applied only to a hypertable that holds no data yet, so upgrading never deletes history: for a contract that already holds data the output logs the add_retention_policy statement to run by hand (ENG-5757)
```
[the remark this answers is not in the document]
Aaron99B: changed that

historian_plugin/output_historian.go
```diff
@@ -305,6 +306,18 @@ func (o *historianOutput) Connect(ctx context.Context) error {
...
 		}
 		cfg.ConnConfig.DefaultQueryExecMode = pgx.QueryExecModeExec // simple protocol (pgbouncer txn pool)
+		// pgx drops server notices unless this is set, and the bootstrap needs them: add_*_policy with
+		// if_not_exists => TRUE reports a policy it declined to change as a notice and returns -1, so
+		// without this the decline reaches no log.
+		cfg.ConnConfig.OnNotice = func(_ *pgconn.PgConn, n *pgconn.Notice) {
+			// WARNING is what add_*_policy raises when if_not_exists makes it decline; NOTICE is
+			// routine bootstrap chatter ("trigger ... does not exist, skipping") on every start.
+			if n.Severity == "WARNING" {
+				o.logger.Warnf("historian: %s", n.Message)
```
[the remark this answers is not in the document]
Aaron99B: changed that

historian_plugin/sql.go
```diff
@@ -303,30 +303,55 @@ func sub(sql string, contract string) string {
...
+func policyJobExistsSQL(procName string, hypertableName string) string {
+	return fmt.Sprintf("EXISTS (SELECT 1 FROM timescaledb_information.jobs WHERE proc_name = '%s' AND hypertable_schema = 'umh' AND hypertable_name = '%s')", procName, hypertableName)
+}
+
+func hypertableHasChunksSQL(hypertableName string) string {
+	return fmt.Sprintf("EXISTS (SELECT 1 FROM timescaledb_information.chunks WHERE hypertable_schema = 'umh' AND hypertable_name = '%s')", hypertableName)
+}
+
+func compressionEnabledSQL(hypertableName string) string {
+	return fmt.Sprintf("EXISTS (SELECT 1 FROM timescaledb_information.hypertables WHERE hypertable_schema = 'umh' AND hypertable_name = '%s' AND compression_enabled)", hypertableName)
```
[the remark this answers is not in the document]
Aaron99B: changed that

historian_plugin/output_historian.go
Aaron99B: I adapted the error logging structure to be more clear

historian_plugin/output_historian.go
Aaron99B: Added the tests

historian_plugin/output_historian.go
Aaron99B: did that

historian_plugin/output_integration_test.go
Aaron99B: unified it

historian_plugin/output_integration_test.go
Aaron99B: did that

historian_plugin/transform.go
```diff
@@ -71,6 +71,13 @@ func NormalizeContract(metaContract string) string {
 
 // ValidateContract checks that data_contract_name is a bare lowercase name (letters, digits,
 // underscores) with no leading underscore and no _vN version suffix.
+// maxContractLen keeps "attribute_" + the contract name inside PostgreSQL's 63-byte identifier
+// limit. Past it the server truncates the table name it creates, and two contracts that differ only
+// in the truncated tail share one pair of hypertables, so one contract's rows, policies and drift
+// warnings become the other's.
+// https://www.postgresql.org/docs/current/limits.html
+const maxContractLen = 63 - len("attribute_")
```
FalconTube, line 79: As discussed, let's bring this check into umh-core as well.
But just a follow up, nothing to here in this PR

FalconTube (approved):
good to go now :)

historian_plugin/transform.go
Aaron99B: https://linear.app/united-manufacturing-hub/issue/ENG-5775/reject-over-long-data-model-names-in-umh-core-not-only-at-the

.changelog/unreleased/fixes-20260826-154709.yaml
```diff
@@ -0,0 +1,3 @@
+kind: fixes
+body: The historian output now applies compression and retention to every data contract, not only the first one on a database; chunks on every later contract were previously never compressed and never dropped. Retention reaches only a hypertable that holds no data yet, so upgrading never deletes stored rows, and for a contract that already holds data the output logs the add_retention_policy statement to run by hand (ENG-5757)
```
JeremyTheocharis: maybe make this shorter? try to explain it to @m4jus?

led0nk (approved):

PR #444, opened 2026-08-27
Drop timestamp_ms from stored historian tag metadata

Aaron99B:
## Problem

The historian stores `timestamp_ms` as tag metadata. Its value is different on every
message, so the attribute fingerprint changes every time and each data point writes a
row into the attribute table.

## What we are shipping

- `timestamp_ms` joins the built-in high-churn exclusion list, so it is no longer stored
  as an attribute in the default `metadata_keys_all` mode.

## What we are NOT shipping

- Nothing cleans up attribute rows already written; existing tables keep their per-point
  rows until pruned by hand.
- Allowlist mode (`metadata_keys_all=false`) is unchanged by design: an explicit
  `metadata_keys` entry still wins, and the existing high-churn warning covers it.

Fixes ENG-5768


FalconTube (approved):
good to go

PR #458, opened 2026-09-23
feat: s7comm configurable pollrate

led0nk:
## Problem

Previously one could not configure the `pollRate` meaning the frequency how the device (plc) gets requested at. Although this is not explicitely needed, this can make sense for specific use-cases, such as "rising edge" or similar things.

## What we are shipping

- option to configure `pollRate` (`time.Duration`)
 -> ends up more as a "waiting" time or as in modbus stated `timeBetweenReads`
- guard against negative or zero-values

## What we are NOT shipping

- refactor with `time.Ticker` to exactly enable a polling-behaviour
- refactor for other protocols to use `time.Duration`

## Tested

- [ ] unit-test(s)
- [ ] live test with benthos-umh running against s7-1200

Fixes ENG-6196


s7comm_plugin/s7comm.go
```diff
@@ -394,18 +421,25 @@ func (g *S7CommInput) ReadBatch(_ context.Context) (service.MessageBatch, servic
...
-func (g *S7CommInput) Close(_ context.Context) error {
-	if g.Handler != nil {
-		g.Handler.Close()
-		g.Handler = nil
-		g.Client = nil
+func (s *S7CommInput) pause(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-time.After(s.PollRate):
```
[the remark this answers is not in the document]
led0nk: we guard against this when parsing the configfields, although this here might be a good option to have it simplified, it only then would face the user whenever already reading out data and not while parsing the config at the first part

s7comm_plugin/s7comm.go
```diff
@@ -314,60 +334,67 @@ func init() {
...
-func (g *S7CommInput) ReadBatch(_ context.Context) (service.MessageBatch, service.AckFunc, error) {
-	if g.Client == nil {
+func (s *S7CommInput) ReadBatch(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
+	if s.Client == nil {
 		return nil, nil, fmt.Errorf("S7Comm client is not initialized")
 	}
 
+	// NOTE: we define this as pollRate whereas in reality it doesn't respect true
+	// read time, it just adds a waiting time before the read.
+	err := s.pause(ctx)
```
JeremyTheocharis: i think in opcua we put this at the end of the readbatch call, maybe we can make it similar as we did it there?

docs/input/siemens-s7.md
```diff
@@ -28,6 +29,7 @@ input:
 * **tcpDevice**: IP address or hostname of the Siemens S7 PLC, optionally with port (e.g., `192.168.0.1:102`). If no port is specified, the default S7 port 102 is used.
 * **rack**: Identifies the physical location of the CPU within the PLC rack.
 * **slot**: Identifies the specific CPU slot within the rack.
+* **pollRate**: Wait between two read cycles, as a duration string such as `500ms` or `2s`. Defaults to `1s`. Low values can overload the PLC. A value of zero or less logs a warning and falls back to `1s`.
```
JeremyTheocharis: maybe check with @robertschwarz-umh?

.changelog/unreleased/improvements-20260924-162001.yaml
```diff
@@ -0,0 +1,3 @@
+kind: improvements
+body: The S7 input takes `pollRate`, a duration such as `500ms` or `2s`, setting how long it waits between reads. It defaults to `1s`, the interval the input was previously fixed at, and a very low rate can overload the PLC (ENG-6196)
```
JeremyTheocharis: maybe make this more user facing? check with @m4jus

JeremyTheocharis (changes requested):

s7comm_plugin/s7comm.go
led0nk: no it should be adjusted in opcua
with `s.pause(ctx)` on top you:
- don't throw away a succesfully read batch if the context cancels while pausing
- in modbus it's the same behaviour because the behaviour is more "timeBetweenReads" than pollRate
- if you error from reading a batch with `service.ErrNotConnected` you will meet  the waiting time not just skipping it

